PR #1, opened 2020-08-31
Adding archived email to gerrit conversion

WLIU0628:
This feature allows for single email threads to be converted to Gerrit. Link to Gerrit: https://linux-review.googlesource.com/c/linux/kernel/git/torvalds/linux/+/2781

### DIFF
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -16,22 +16,19 @@ import email
 from setup_gmail import Message
 
 def generate_email_from_file(file: str):
-    raw_email = open(file, "r")
-    compiled_email = email.message_from_string(raw_email.read())
-    raw_email.close()
-    return email_to_message(compiled_email)
+    with open(file, "r") as raw_email:
+        compiled_email = email.message_from_string(raw_email.read())
+        return _email_to_message(compiled_email)
     
-def email_to_message(compiled_email) -> Message:
+def _email_to_message(compiled_email) -> Message:
     content = []
     if compiled_email.is_multipart():
         for payload in compiled_email.get_payload():
             content.append(payload.get_payload())
     else:
         content = compiled_email.get_payload()
-    return Message(compiled_email['Message-Id'], compiled_email['subject'], compiled_email['from'], compiled_email['In-Reply-To'], content)
-
-def main():
-    generate_email_from_patches("test_data/patch6.txt")
-    
-if __name__ == "__main__":
-    main()
+    return Message(compiled_email['Message-Id'],
+                   compiled_email['subject'],
+                   compiled_email['from'],
+                   compiled_email['In-Reply-To'],
+                   content)
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -1,0 +1,37 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import email
+from setup_gmail import Message
+
+def generate_email_from_file(file: str):
+    raw_email = open(file, "r")
+    compiled_email = email.message_from_string(raw_email.read())
+    raw_email.close()
+    return email_to_message(compiled_email)
+    
+def email_to_message(compiled_email) -> Message:
+    content = []
+    if compiled_email.is_multipart():
+        for payload in compiled_email.get_payload():
+            content.append(payload.get_payload())
+    else:
+        content = compiled_email.get_payload()
+    return Message(compiled_email['Message-Id'], compiled_email['subject'], compiled_email['from'], compiled_email['In-Reply-To'], content)
+
+def main():
+    generate_email_from_patches("test_data/patch6.txt")
+    
+if __name__ == "__main__":
+    main()
--- a/gerrit.py
+++ b/gerrit.py
@@ -29,6 +29,7 @@ from requests.auth import AuthBase
 from http.cookiejar import CookieJar, MozillaCookieJar
 from setup_gmail import Message
 from setup_gmail import find_thread
+from archive_converter import generate_email_from_file
 
 def get_gerrit_rest_api(cookie_jar_path: str, gerrit_url: str) -> GerritRestAPI:
     cookie_jar = MozillaCookieJar(cookie_jar_path)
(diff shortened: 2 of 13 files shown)


archive_converter.py
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import email
+from setup_gmail import Message
+
+def generate_email_from_file():
+    raw_email = open("patches/patch6.txt", "r")
```
bjh83: Can we make generate_email_from_file() take the file path as a parameter?

gerrit.py
```diff
@@ -120,7 +121,7 @@ def main():
     gerrit_git = GerritGit(git_dir='gerrit_git_dir',
```
bjh83, line 121: How about we break all of this code in the main() function out into a test? (Since that is basically what it is.)

patch_parser.py
```diff
@@ -348,12 +350,17 @@ def parse_comments(email_thread: Message) -> Patchset:
         comments = []
         for reply in patch.children:
             comments.extend(diff_reply(patch, reply))
-        set_index, length = parse_set_index(patch)
-        assert length == len(patches)
-        text = 'From: {from_}\nSubject: {subject}\n\n{content}'.format(
-                from_=patch.from_, subject=patch.subject, content=patch.content)
-        patch_list.append(Patch(text=patch.content, text_with_headers=text, set_index=set_index, comments=comments))
-        patch_list.sort(key=lambda x: x.set_index)
+        if (email_thread.in_reply_to):
```
bjh83: What does this new block of code do?

patches/patch4.txt
```diff
@@ -0,0 +1,39 @@
+Return-Path: <linux-kernel-owner+w=401wt.eu-S1751045AbXAVR6a@vger.kernel.org>
```
bjh83: Can we rename the patches/ directory to test_data?

bjh83 (changes requested):
Overall, I think this is pretty good.

One thing about the commit structure: Can you break this up into logical commits? Like take all the fixup commits and squash them into the feature commits?

Basically, I like when the commit history reflects the logical changes made to the project.

patch_parser.py
WLIU0628: I realize now that it doesn't handle cover letters anymore. But, I was trying to handle single email threads (with only the patch being sent).

gerrit.py
WLIU0628: any resources on starting tests for Python?

archive_converter.py
WLIU0628: yeah that was a thought that crossed my mind after I submitted the PR haha

patches/patch4.txt
WLIU0628: yessir!

archive_converter.py
```diff
@@ -0,0 +1,37 @@
...
+def email_to_message(compiled_email) -> Message:
+    content = []
+    if compiled_email.is_multipart():
+        for payload in compiled_email.get_payload():
+            content.append(payload.get_payload())
+    else:
+        content = compiled_email.get_payload()
+    return Message(compiled_email['Message-Id'], compiled_email['subject'], compiled_email['from'], compiled_email['In-Reply-To'], content)
+
+def main():
```
bjh83: I think this main function is unnecessary because of the test you wrote.

gerrit.py
bjh83: Given the test you wrote above, do you still have any questions about this?

bjh83 (changes requested):
> This feature allows for single email threads to be converted to Gerrit. Link to Gerrit: https://linux-review.googlesource.com/c/linux/kernel/git/torvalds/linux/+/2781

Can you add this to your commit message below the subject line? Also mention that you added tests.

archive_converter.py
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+
+import email
+from setup_gmail import Message
+
+def generate_email_from_file(file: str):
+    raw_email = open(file, "r")
+    compiled_email = email.message_from_string(raw_email.read())
+    raw_email.close()
+    return email_to_message(compiled_email)
```
dlatypov: Can we simplify this to:
```suggestion
    with open(file, "r") as raw_email:
        compiled_email = email.message_from_string(raw_email.read())
        return email_to_message(compiled_email)
```

archive_converter.py
```diff
@@ -0,0 +1,37 @@
...
+    
+def email_to_message(compiled_email) -> Message:
```
```suggestion
def _email_to_message(compiled_email) -> Message:
```

I don't imagine anyone will be using this func outside of this file, so let's make this "private"
We'd probably just have the file based one for testing and a helper that more directly fetches from the archive.

patch_parser.py
dlatypov: Might be a good idea to add a small unit test for this function.

patch_parser.py
```diff
@@ -348,12 +350,17 @@ def parse_comments(email_thread: Message) -> Patchset:
...
+            set_index, length = parse_set_index(patch)
+            assert length == len(patches)
+            text = 'From: {from_}\nSubject: {subject}\n\n{content}'.format(
+                    from_=patch.from_, subject=patch.subject, content=patch.content)
+            patch_list.append(Patch(text=patch.content, text_with_headers=text, set_index=set_index, comments=comments))
+            patch_list.sort(key=lambda x: x.set_index)
+        else:
```
dlatypov: seems like this branch only differs in terms of set_index.

Perhaps we can refactor this to

```py
# Patchsets with a single patch might not include "PATCH 1/1"
  if len(patches) == 1:
    set_index = 0
  else:
    set_index, length = parse_set_index(patch)
    assert length == len(patches)
```

dlatypov (changes requested):

patch_parser.py
WLIU0628: so I wasn't sure how complex everything was behind the scenes, which is why I used the else instead of just adjusting the set index and length. But, yeah I definitely agree this is the way to go.

archive_converter.py
```diff
@@ -16,22 +16,15 @@
...
         content = compiled_email.get_payload()
-    return Message(compiled_email['Message-Id'], compiled_email['subject'], compiled_email['from'], compiled_email['In-Reply-To'], content)
-
-def main():
-    generate_email_from_patches("test_data/patch6.txt")
-    
-if __name__ == "__main__":
-    main()
\ No newline at end of file
+    return Message(compiled_email['Message-Id'], compiled_email['subject'], compiled_email['from'], compiled_email['In-Reply-To'], content)
```
bjh83: Nit: Can you limit your line length to 80 chars?

patch_parser_test.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+from patch_parser import parse_comments
+from archive_converter import generate_email_from_file
+
+class Patch_Parser_Test(unittest.TestCase):
+
+    def test_parse_comments_for_single_email_thread(self):
```
bjh83, line 7: Can you also add test cases for:

1. Multiple patches, no cover letter.
2. Multiple patches with cover letter.
3. Email no patches.

bjh83 (commented):
Looking good! Two minor suggestions...

patch_parser_test.py
bjh83: We talked offline and I agree that the other test cases I asked for will not work at this point since there is no "index" to build up email threads.

Since adding the index is outside the scope of this change, I am fine with just putting a TODO comment here as a reminder to add the other test cases when possible.

bjh83 (approved):

dlatypov (approved):

PR #2, opened 2020-09-02
Fix all pre-existing python typing issues.

dlatypov:
Type annotations are present in the code, but these are not checked at
runtime, see https://docs.python.org/3/library/typing.html

This doesn't get it clean, but it does fix most of the current issues.

$ mypy --ignore-missing-imports *.py
Before: Found 44 errors in 4 files (checked 4 source files)
After:  Found 13 errors in 2 files (checked 4 source files)

Fix some of the simpler issues, e.g. multiple-return is expressed via `Tuple[a,b]`, not `(a,b)`.

Also avoid reusing variables for values of different types, which
tripped up the type checker, e.g. reusing `line` for a `Line` and a `List[str]`.

### DIFF
--- a/setup_gmail.py
+++ b/setup_gmail.py
@@ -17,6 +17,7 @@ import base64
 import pickle
 import os.path
 import re
+from typing import Optional
 from googleapiclient.discovery import build
 from google_auth_oauthlib.flow import InstalledAppFlow
 from google.auth.transport.requests import Request
@@ -38,7 +39,7 @@ class MessageKey(object):
 
 class PatchsetFetcher(object):
     def fetch_patchset(self, key: MessageKey) -> Patchset:
-        pass
+        raise NotImplementedError()
 
 class Thread(object):
     def __init__(self, key: MessageKey, message_ids=[]):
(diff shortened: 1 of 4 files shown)


dlatypov:
Went ahead and fixed up the remaining type errors since they were simpler than I had originally thought.
Also noticed that the code was currently printing to stdout and returning None to signal errors that it couldn't recover from, so raising an exception fixed type errors and made the code a bit more debuggable.

```
$ mypy --ignore-missing-imports *.py
Success: no issues found in 4 source files
```

bjh83:
LGTM.

setup_gmail.py
```diff
@@ -190,7 +191,7 @@ def get_service():
 
 class GmailPatchsetFetcher(object):
```
bjh83, line 192: I don't think this class is used anywhere. You could probably just remove it.

bjh83 (commented):
I left one minor comment. Feel free to address it/merge it at your own leisure.

setup_gmail.py
dlatypov: Yeah, it's not used.
I was debating removing it, but deleting code felt a bit out-of-scope for these patches.

I'll submit as-is for now.

PR #3, opened 2020-09-11
Adding multi email gerrit conversion

WLIU0628:
now supports multi email thread to Gerrit conversion as seen here: https://linux-review.googlesource.com/c/linux/kernel/git/torvalds/linux/+/2861
also added tests for multi email thread conversion

### DIFF
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -13,13 +13,72 @@
 # limitations under the License.
 
 import email
+import os
+from typing import List, Dict
 from setup_gmail import Message
+
+class ArchiveMessageIndex(object):
+    def __init__(self):
+        # Maps message subject (str) to starting message of email thread
+        self._key_to_thread_start : Dict[str, Message] = {}
+        # Maps message.id to message
+        self._messages_seen : Dict[str, Message] = {}
+
+    def find_thread(self, key: str) -> Message:
+        return self._key_to_thread_start[key]
+
+    def update(self, data_dir: str):
+        new_messages : List[Message] = []
+
+        for filename in os.listdir(data_dir):
+            if not filename.endswith(".txt"):
+                continue
+            email = generate_email_from_file(os.path.join(data_dir, filename))
+            if email.id not in self._messages_seen:
+                new_messages.append(email)
+        self._add_messages_to_index(new_messages)
+
+    def size(self):
+        return len(self._messages_seen)
+
+    def _add_messages_to_index(self, new_messages : List[Message]):
+        """ Iterates through all new emails and links together emails that form a thread by populating message.children. """
+
+        need_parent : List[Message] = []
+
+        # First iterates through all messages to distinguish between 1.) replies and 2.) the start of a thread.
+        for message in new_messages:
+            if message.id not in self._messages_seen:
+                self._messages_seen[message.id] = message
+
+            # If message is a reply, it needs to be associated with an existing thread
+            if message.in_reply_to is not None:
+                need_parent.append(message)
+                continue
+
+            # If message doesn't have the LKML style [PATCH v4 00/11],
+            # we don't need to put it in the index.
+            key = message.get_key()
+            if key:
+                self._key_to_thread_start[key] = message
+
+        for message in need_parent:
+            if message.in_reply_to not in self._messages_seen:
+                print("Could not find parent email, dropping " + message.subject)
+                continue
+            parent = self._messages_seen[message.in_reply_to]
+            parent.children.append(message)
+
+def find_thread(key: str) -> Message:
+    archive_index = ArchiveMessageIndex()
+    archive_index.update('test_data')
+    return archive_index.find_thread(key)
 
 def generate_email_from_file(file: str):
     with open(file, "r") as raw_email:
         compiled_email = email.message_from_string(raw_email.read())
         return _email_to_message(compiled_email)
-    
+
 def _email_to_message(compiled_email) -> Message:
     content = []
     if compiled_email.is_multipart():
--- a/archive_converter_test.py
+++ b/archive_converter_test.py
@@ -1,7 +1,7 @@
 import unittest
-from archive_converter import generate_email_from_file
+from archive_converter import generate_email_from_file, ArchiveMessageIndex
 
-class Archive_Converter_Test(unittest.TestCase):
+class ArchiveConverterTest(unittest.TestCase):
 
     def test_generate_email_from_single_email_thread(self):
         email = generate_email_from_file("test_data/patch6.txt")
@@ -9,7 +9,15 @@ class Archive_Converter_Test(unittest.TestCase):
         self.assertEqual(email.in_reply_to, None)
         self.assertEqual(email.from_, "Raul E Rangel <rrangel@chromium.org>")
         self.assertTrue(len(email.content) > 0)
-        
+
+    def test_update_with_no_changes_to_data(self):
+        archive_index = ArchiveMessageIndex()
+        archive_index.update('test_data')
+        old_size = archive_index.size()
+        archive_index.update('test_data')
+        self.assertEqual(old_size, archive_index.size())
+
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/gerrit.py
+++ b/gerrit.py
@@ -28,8 +28,7 @@ from requests import PreparedRequest
 from requests.auth import AuthBase
 from http.cookiejar import CookieJar, MozillaCookieJar
 from setup_gmail import Message
-from setup_gmail import find_thread
-from archive_converter import generate_email_from_file
+from archive_converter import find_thread
 
 def get_gerrit_rest_api(cookie_jar_path: str, gerrit_url: str) -> GerritRestAPI:
     cookie_jar = MozillaCookieJar(cookie_jar_path)
@@ -123,7 +122,7 @@ def main():
     gerrit_git = GerritGit(git_dir='gerrit_git_dir',
                            cookie_jar_path='gerritcookies',
                            url=gob_url, project='linux/kernel/git/torvalds/linux', branch='master')
-    email_thread = generate_email_from_file("test_data/patch6.txt")
+    email_thread = find_thread('PATCH v2 0/4')
     patchset = parse_comments(email_thread)
     gerrit_git.apply_patchset_and_cleanup(patchset)
     find_and_label_all_revision_ids(gerrit, patchset)
--- a/patch_parser.py
+++ b/patch_parser.py
@@ -314,7 +314,7 @@ def diff_reply(parent: Message, child: Message) -> List[Comment]:
 def filter_patches_and_cover_letter_replies(email_thread: Message) -> Tuple[List[Message], List[Message]]:
     patches = []
     cover_letter_replies = []
-    if (not email_thread.in_reply_to):
+    if (not email_thread.in_reply_to and parse_set_index(email_thread)[0] == 1):
         patches.append(email_thread)
     for message in email_thread.children:
         if re.match(r'\[.+ \d+/\d+\] .+', message.subject):
--- a/patch_parser_test.py
+++ b/patch_parser_test.py
@@ -1,20 +1,37 @@
 import unittest
 from patch_parser import parse_comments
-from archive_converter import generate_email_from_file
+from archive_converter import find_thread, generate_email_from_file
 
-class Patch_Parser_Test(unittest.TestCase):
+class PatchParserTest(unittest.TestCase):
 
     def test_parse_comments_for_single_email_thread(self):
-        email = generate_email_from_file("test_data/patch6.txt")
-        patchset = parse_comments(email)
+        patchset = parse_comments(find_thread('PATCH v2 1/2'))
+
         self.assertTrue(len(patchset.patches) > 0)
-        self.assertTrue(patchset.patches[0].set_index == 0)
-        self.assertTrue(len(patchset.patches[0].text) > 0)
-        self.assertTrue(len(patchset.patches[0].text_with_headers) > 18)
-        self.assertEqual(patchset.patches[0].comments, [])
-        
-    #TODO(willliu@google.com): Add tests for Multiple patches, no cover letter, Multiple patches with cover letter, Email no patches.
-        
+        first_patch = patchset.patches[0]
+        self.assertEqual(first_patch.set_index, 0)
+        self.assertNotEqual(first_patch.text, '')
+        self.assertIn('[PATCH v2 1/2] Input: i8042 - Prevent intermixing i8042 commands', first_patch.text_with_headers)
+        self.assertEqual(first_patch.comments, [])
+
+    def test_parse_comments_for_multi_email_thread_with_cover_letter(self):
+        patchset = parse_comments(find_thread('PATCH v2 0/4'))
+
+        self.assertEqual(len(patchset.patches), 4)
+        first_patch = patchset.patches[0]
+        self.assertEqual(first_patch.set_index, 1)
+        self.assertIn('[PATCH v2 1/4] kselftests/arm64: add a basic Pointer Authentication test', first_patch.text_with_headers)
+        self.assertNotEqual(first_patch.text, '')
+        self.assertEqual(first_patch.comments, [])
+        self.assertIn('[PATCH v2 2/4] kselftests/arm64: add nop checks for PAuth tests', patchset.patches[1].text_with_headers)
+
+    def test_parse_comments_for_non_patch_email(self):
+        patchset = parse_comments(generate_email_from_file('test_data/thread_patch0.txt'))
+
+        self.assertEqual(len(patchset.patches), 0)
+
+    #TODO(willliu@google.com): Add tests for Multiple patches, no cover letter
+
 
 if __name__ == '__main__':
     unittest.main()


patch_parser_test.py
```diff
@@ -0,0 +1,18 @@
...
+from patch_parser import parse_comments
+from archive_converter import generate_email_from_file
+
+class Patch_Parser_Test(unittest.TestCase):
+
+    def test_parse_comments_for_single_email_thread(self):
+        email = generate_email_from_file("test_data/patch6.txt")
+        patchset = parse_comments(email)
+        self.assertTrue(len(patchset.patches) > 0)
+        self.assertTrue(patchset.patches[0].set_index == 0)
```
dlatypov:
```suggestion
        self.assertEqual(patchset.patches[0].set_index, 0)
```

patch_parser_test.py
```diff
@@ -0,0 +1,18 @@
...
+        self.assertTrue(len(patchset.patches[0].text) > 0)
+        self.assertTrue(len(patchset.patches[0].text_with_headers) > 18)
```
dlatypov:
```suggestion
        self.assertGreater(len(patchset.patches[0].text_with_headers), 18)
```

This way it'll show the actual length.

Note: I didn't suggest this for all previous checks bc they were `> 0` and it'd be obvious what the length we got was.

Ditto these comments for the child commit.

archive_converter.py
```diff
@@ -13,8 +13,57 @@
 # limitations under the License.
 
 import email
+import os
 from setup_gmail import Message
 
+class ArchiveMessageIndex(object):
+    def __init__(self):
+        self._lookup_index = {} # Maps MessageKey to Thread
```
dlatypov: add typing info here

```py
self._lookup_index : Dict[str, Message] = {}
```

But actually, I think a name change could be useful.
We generally "lookup" things in an "index", so the name seems a bit redundant atm, given it's not returning an "index" but a Message (according to type annotation below)

```py
self._key_to_thread_start : Dict[str, Message] = {}
```

perhaps?

dlatypov: It says "MessageKey" in the comment, but based on `find_thread`'s type annotation, I assume the key is `str`?

There is a "MessageKey" type, so I'm not sure which one of these to trust.
Can you update the comment (or using tying to make it clear?)

archive_converter.py
```diff
@@ -13,8 +13,57 @@
...
+                    self._lookup_index[key] = message
+                else:
+                    # If message doesn't have the LKML style [PATCH v4 00/11],
+                    # we don't need to put it in the index.
+                    continue
+            else:
+                need_parent.append(message)
+        for message in need_parent:
+            if message.in_reply_to not in self._messages_seen:
+                # If we don't have the parent message, then we can't reconstruct
```
dlatypov: Hmm, seems like we should maybe print something here to signal this rather than silently drop it.

This code isn't using https://docs.python.org/3/library/logging.html, so perhaps just a print statement would be fine.

archive_converter.py
```diff
@@ -13,8 +13,57 @@
...
+        self._messages_seen = {} # Maps message.id to message
+
+    def find_thread(self, key: str) -> Message:
+        return self._lookup_index[key]
+
+    def update(self):
+        fully_updated = False
+        new_messages = []
+
+        for filename in os.listdir('test_data'):
```
dlatypov: can we pass the dir in as a parameter?

Also, we'd want to us `os.path.join(dir, filename)` to produce the child filepath below

archive_converter.py
```diff
@@ -13,8 +13,57 @@
...
+                need_parent.append(message)
+        for message in need_parent:
+            if message.in_reply_to not in self._messages_seen:
+                # If we don't have the parent message, then we can't reconstruct
+                # the patchset for this group of emails anyway, so just skip.
+                continue
+            parent = self._messages_seen[message.in_reply_to]
+            parent.children.append(message)
+            
+def find_thread(key: str) -> Message:
```
dlatypov, line 72: I assume this function is more for convenience and that the users of this module are going to make index objects and update them manually?
(rather than calling this func and recreating the index every time)

If so, do you want to move it into the test file as a helper there?

archive_converter.py
```diff
@@ -13,8 +13,57 @@
...
+    def __init__(self):
+        self._lookup_index = {} # Maps MessageKey to Thread
+        self._messages_seen = {} # Maps message.id to message
+
+    def find_thread(self, key: str) -> Message:
+        return self._lookup_index[key]
+
+    def update(self):
+        fully_updated = False
+        new_messages = []
```
```suggestion
        new_messages : List[Message] = []
```

and similarly annotate `_add_messages_to_index()`

archive_converter.py
```diff
@@ -13,8 +13,57 @@
...
+
+        for filename in os.listdir('test_data'):
+            if filename.endswith(".txt"):
+                email = generate_email_from_file('test_data/' + filename)
+                if(email.id not in self._messages_seen):
+                    new_messages.append(email)
+        self._add_messages_to_index(new_messages)
+
+    def _add_messages_to_index(self, new_messages):
```
dlatypov: Mind adding a docstring for this function as well?

Would be good to document the pre/post-conditions of this func.
E.g. one of the key responsibilities is that it populates `Message.children`

archive_converter.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def update(self):
+        fully_updated = False
+        new_messages = []
+
+        for filename in os.listdir('test_data'):
+            if filename.endswith(".txt"):
+                email = generate_email_from_file('test_data/' + filename)
+                if(email.id not in self._messages_seen):
+                    new_messages.append(email)
```
dlatypov: nit: reduce nesting
And by that we mean, prefer to return/continue early rather than nesting `if`s deeper and deeper

```suggestion
            if not filename.endswith(".txt"):
                continue
            email = generate_email_from_file('test_data/' + filename)
            if(email.id not in self._messages_seen):
                new_messages.append(email)
```

It's a bit easier to read since we don't not have to keep in mind how deep we are.
Readers can think "oh, I can assume it's a .txt file for the lines after this" instead of "it's a .txt file in this nested block"

archive_converter.py
```diff
@@ -0,0 +1,83 @@
...
+        self._add_messages_to_index(new_messages)
+
+    def _add_messages_to_index(self, new_messages):
+        need_parent = []
+        for message in new_messages:
```
dlatypov, line 50: I'd add a comment above this loop, something like

```
# First find all the root nodes (brainfart: can't think of the more appropriate name in this context)
```

archive_converter_test.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+from archive_converter import generate_email_from_file
+
+class Archive_Converter_Test(unittest.TestCase):
+
```
dlatypov, line 5: Can we add a test that calling `update()` on the index is idempotent?

E.g.
```python
index.update('test_data')
index.update('test_data')  # this should do no additional new work
```

archive_converter_test.py
```diff
@@ -0,0 +1,15 @@
+import unittest
+from archive_converter import generate_email_from_file
+
+class Archive_Converter_Test(unittest.TestCase):
```
dlatypov: ArchiveConverterTest is fine

archive_converter_test.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def test_generate_email_from_single_email_thread(self):
+        email = generate_email_from_file("test_data/patch6.txt")
+        self.assertEqual(email.subject, '[PATCH v2 1/2] Input: i8042 - Prevent intermixing i8042 commands')
+        self.assertEqual(email.in_reply_to, None)
+        self.assertEqual(email.from_, "Raul E Rangel <rrangel@chromium.org>")
+        self.assertTrue(len(email.content) > 0)
+
```
dlatypov: can we have a test here that checks we map the parent to the right child?

E.g. calls `update()` and then we check that the subject of the child matches what we want, for example?
(Don't need it to check too deeply)

archive_converter.py
```diff
@@ -0,0 +1,83 @@
...
+
+            if message.in_reply_to is None:
+                key = message.get_key()
+                print(key)
+                if key:
+                    self._lookup_index[key] = message
+                else:
+                    # If message doesn't have the LKML style [PATCH v4 00/11],
+                    # we don't need to put it in the index.
+                    continue
```
dlatypov: Ditto, put this continue earlier

```py
if key is None:
    # comment
    continue
...
```

I'd also flip the outer conditional as well

```py
if messsage.in_reply_to is not None:
     need_parent.append(message)
     continue
...
```

gerrit.py
```diff
@@ -29,6 +29,7 @@
 from http.cookiejar import CookieJar, MozillaCookieJar
 from setup_gmail import Message
 from setup_gmail import find_thread
```
dlatypov: did you want to remove the other "find_thread" import?

patch_parser.py
```diff
@@ -316,6 +316,8 @@ def diff_reply(parent: Message, child: Message) -> List[Comment]:
 def filter_patches_and_cover_letter_replies(email_thread: Message) -> (List[Message], List[Message]):
     patches = []
     cover_letter_replies = []
+    if (not email_thread.in_reply_to and re.match(r'\[.+ 1/\d+\] .+', email_thread.subject)):
```
dlatypov: would using `parse_set_index(email_thread)` be cleaner here?
We'd just check that the first value is 1.

dlatypov (changes requested):

archive_converter.py
WLIU0628: yes! also, finding specific threads within the index will probably become deprecated once the application is running on a server. I'll move it to the test file!

bjh83:
Would you mind rebasing this change on master (given that we just merged your preceding pull request)?

archive_converter.py
WLIU0628: Actually, I've thought about it a bit more. Does the location matter? The setup_gmail version of find_thread was defined in the exact same way. It just feels wrong to import find_thread from a test file, because gerrit.py's main function still uses find_thread.

archive_converter.py
```diff
@@ -13,25 +13,83 @@
...
+            email = generate_email_from_file(os.path.join(data_dir, filename))
+            if email.id not in self._messages_seen:
+                new_messages.append(email)
+        self._add_messages_to_index(new_messages)
+        
+    def size(self):
+        return len(self._messages_seen)
+
+    def _add_messages_to_index(self, new_messages : List[Message]):
+        ''' Iterates through all new emails and links together emails that form a thread by populating message.children.
```
dlatypov: I'm more use to this looking like
```python
"""Iterates..."""
```
https://www.python.org/dev/peps/pep-0257

archive_converter.py
```diff
@@ -13,8 +13,69 @@
...
+        self._add_messages_to_index(new_messages)
+        
+    def size(self):
+        return len(self._messages_seen)
+
+    def _add_messages_to_index(self, new_messages : List[Message]):
+        ''' Iterates through all new emails and links together emails that form a thread by populating message.children. 
+        
+            ARGS:
+            new_messages::List[Message] - a list of messages that have yet to be added to the index
```
dlatypov: Think we can drop the Args section.
The var name is clear enough, IMO.

archive_converter.py
```diff
@@ -13,8 +13,69 @@
...
+                
+            # If message doesn't have the LKML style [PATCH v4 00/11],
+            # we don't need to put it in the index.
+            key = message.get_key()
+            if key:
+                self._key_to_thread_start[key] = message
+                
+        for message in need_parent:
+            if message.in_reply_to not in self._messages_seen:
+                print("No Parent Message for " + message.subject)
```
dlatypov: nit: let's make this print be more clear what the implications are.
```python
'Could not find parent email, dropping ' + message.subject
```

Might be value in showing the offending `in_reply_to`, but this would be fine for now, imo

archive_converter_test.py
```diff
@@ -10,6 +10,13 @@ def test_generate_email_from_single_email_thread(self):
         self.assertEqual(email.from_, "Raul E Rangel <rrangel@chromium.org>")
         self.assertTrue(len(email.content) > 0)
         
+    def test_update(self):
+        archive_index = ArchiveMessageIndex()
+        archive_index.update('test_data')
+        size = archive_index.size()
+        self.assertEqual(size, archive_index.size())
```
dlatypov:
```suggestion
        old_size = archive_index.size()
        archive_index.update('test_data')
        self.assertEqual(archive_index.size(), old_size)
```

Perhaps also rename this test case to be more clear what it's testing

patch_parser_test.py
```diff
@@ -1,19 +1,38 @@
...
         patchset = parse_comments(email)
         self.assertTrue(len(patchset.patches) > 0)
-        self.assertTrue(patchset.patches[0].set_index == 0)
+        self.assertEqual(patchset.patches[0].set_index, 0)
         self.assertTrue(len(patchset.patches[0].text) > 0)
-        self.assertTrue(len(patchset.patches[0].text_with_headers) > 18)
+        self.assertGreater(len(patchset.patches[0].text_with_headers),18)
         self.assertEqual(patchset.patches[0].comments, [])
         
+<<<<<<< HEAD
```
dlatypov: looks like you accidentally left in a conflict resolution marker?

patch_parser_test.py
```diff
@@ -1,19 +1,38 @@
...
     #TODO(willliu@google.com): Add tests for Multiple patches, no cover letter, Multiple patches with cover letter, Email no patches.
+=======
+    def test_parse_comments_for_multi_email_thread_with_cover_letter(self):
+        email = find_thread('PATCH v2 0/4')
+        patchset = parse_comments(email)
+        self.assertEqual(len(patchset.patches), 4)
+        self.assertEqual(patchset.patches[0].set_index, 1)
+        self.assertTrue(len(patchset.patches[0].text) > 0)
```
dlatypov: .text is a str, right?

So perhaps
```suggestion
        self.assertNotEqual(patchset.patches[0].text, '')
```

or you could do `self.AssertIn(..., 'some substring')`

patch_parser_test.py
```diff
@@ -1,19 +1,38 @@
...
+        self.assertTrue('[PATCH v2 1/4] kselftests/arm64: add a basic Pointer Authentication test' in patchset.patches[0].text_with_headers)
```
dlatypov: self.assertIn

I'd also move this above the .text check
This is the assertion that identifies the patch we're expecting, so it'd be nice if it came first.

patch_parser_test.py
dlatypov: after the len assert, perhaps can do

```python
  first_patch = patchset.patches[0]
```

to reduce syntactic noise a bit.

patch_parser_test.py
```diff
@@ -1,19 +1,38 @@
...
         self.assertTrue(len(patchset.patches[0].text) > 0)
-        self.assertTrue(len(patchset.patches[0].text_with_headers) > 18)
+        self.assertGreater(len(patchset.patches[0].text_with_headers),18)
         self.assertEqual(patchset.patches[0].comments, [])
         
+<<<<<<< HEAD
     #TODO(willliu@google.com): Add tests for Multiple patches, no cover letter, Multiple patches with cover letter, Email no patches.
+=======
+    def test_parse_comments_for_multi_email_thread_with_cover_letter(self):
+        email = find_thread('PATCH v2 0/4')
```
dlatypov: I'd put a line break between this an the asserts.

And perhaps do

```python
  patchset = parse_comments(find_thread('...')

  self.assert...
```

as we don't use `email` anywhere else.

archive_converter.py
```diff
@@ -13,8 +13,69 @@
...
+class ArchiveMessageIndex(object):
+    def __init__(self):
+        self._key_to_thread_start : Dict[str, Message] = {} #Maps message subject (str) to starting message of email thread
+        self._messages_seen : Dict[str, Message] = {} # Maps message.id to message
+ 
+    def find_thread(self, key: str) -> Message:
+        return self._key_to_thread_start[key]
+
+    def update(self, data_dir: str):
+        fully_updated = False
```
dlatypov: this var seems to be no longer used

dlatypov (changes requested):
Looks like you have merge conflict markers left in this version.
Mind taking a look and removing them?

archive_converter.py
bjh83: Agreed. You should not import form a test file. Nevertheless, feel free to pull non-Gmail specific stuff out of the setup_gmail file; a lot of stuff was put there that probably doesn't belong there.

archive_converter.py
```diff
@@ -13,8 +13,65 @@
 # limitations under the License.
 
 import email
+import os
+from typing import List, Dict
 from setup_gmail import Message
 
+class ArchiveMessageIndex(object):
+    def __init__(self):
+        self._key_to_thread_start : Dict[str, Message] = {} #Maps message subject (str) to starting message of email thread
```
bjh83: I think the comment pushes this line over 80 char, if so, please put the comment on the preceding line.

bjh83 (approved):

dlatypov (approved):

archive_converter.py
```diff
@@ -13,8 +13,67 @@
 # limitations under the License.
 
 import email
+import os
+from typing import List, Dict
 from setup_gmail import Message
 
+class ArchiveMessageIndex(object):
+    def __init__(self):
+        #Maps message subject (str) to starting message of email thread
```
```suggestion
        # Maps message subject (str) to starting message of email thread
```

dlatypov: Note: while you're doing this, can you delete the trailing whitespace added in this file?

There are a number of instances in this commit.
E.g. the line right below this.

```sh
$ egrep '\s+$' *.py -c
archive_converter.py:11
archive_converter_test.py:3
gerrit.py:0
git.py:0
patch_parser.py:0
patch_parser_test.py:7
setup_gmail.py:0
```

Can do so by opening up a shell and running
```
$ ls *.py | xargs sed -i -E 's/\s+$//'
```

(there are a few instances from the previous PR which will get changed too, but I'm fine with that)

PR #4, opened 2020-09-22
Git update and archive to file converter

WLIU0628:
added an archive updater which will automatically update the lkml archive and fill a specified directory with the new git commits

### DIFF
--- a/archive_updater.py
+++ b/archive_updater.py
@@ -1,0 +1,62 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import os
+
+# TODO(willliu@google.com): bump this value up
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+    
+def fill_message_directory(archive_path: str, directory: str, last_used_commit_hash: str) -> str:
+    '''Updates the git repo, then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored
+    in the directory corresponding to the passed in directory path. 
+    
+    Args:
+        archive_path: path to the email archive
+        directory: where to store files
+        last_used_commit_hash: ending point for filling the directory
+    
+    Returns:
+        The most recent processed commit hash
+    
+    Raises:
+        CalledProcessError: when git fetch, git log, or git show fails
+        Exception: when the log from git log shows no hashes
+    '''
+    
+    subprocess.check_call(['git', '-C', archive_path, 'fetch'])
+    
+    output = subprocess.check_output(
+        ['git', '-C', archive_path, 'log', '--format=format:%H', '-n',str(MAX_NUMBER_OF_RECENT_COMMITS)])
+    message_hashes = output.decode('utf-8').split()
+    
+    if len(message_hashes) == 0:
+        raise Exception(f'There are no commits in git repo: {archive_path}')
+    
+    for hash in message_hashes:
+        if hash == last_used_commit_hash:
+            break
+        file = os.path.join(directory, f'{hash}.txt')
+        # TODO(willliu@google.com): fetch the message contents on demand. We also don't check for errors creating the file
+        with open(file, 'w') as f:
+            subprocess.call(['git', '-C', archive_path, 'show', f'{hash}:m'],
+                            stdout=f)
+        
+    return message_hashes[0]
+    
+def main():
+    print(fill_message_directory('../linux-kselftest/git/0.git', '../lkml-gerrit-bridge/test_data', '5227e860278407173cbfdab594676764153c27e1'))
+    
+if __name__ == '__main__':
+    main()
--- a/archive_updater_test.py
+++ b/archive_updater_test.py
@@ -1,0 +1,68 @@
+import unittest
+import subprocess
+import os
+import tempfile
+import shutil
+from archive_updater import fill_message_directory, MAX_NUMBER_OF_RECENT_COMMITS
+from unittest import mock
+
+class ArchiveUpdaterFillMessageDirectoryTest(unittest.TestCase):
+
+    def setUp(self):
+        self.tmp_dir = tempfile.mkdtemp()
+    def tearDown(self):
+        shutil.rmtree(self.tmp_dir)
+
+    @mock.patch.object(subprocess, 'check_call')
+    def test_update_failure(self, mock_check_call):
+        mock_check_call.side_effect = subprocess.CalledProcessError(returncode=1,
+                                                                    cmd=['git', '-C', 'archive_path', 'fetch'],
+                                                                    stderr=b'Failed to fetch')
+        with self.assertRaises(subprocess.CalledProcessError):
+            fill_message_directory('archive_path', self.tmp_dir, '')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'call')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_success(self, mock_check_call, mock_call, mock_check_output):
+        def writeFile(args, stdout):
+            stdout.write('called with: ' + ' '.join(args))
+        mock_call.side_effect = writeFile
+        mock_check_output.return_value = b'success1 success2 success3'
+
+        last_used_hash = fill_message_directory('archive_path', self.tmp_dir, '')
+
+        self.assertEqual(len(os.listdir(self.tmp_dir)), 3)
+        for filename in os.listdir(self.tmp_dir):
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertIn('git -C archive_path show', val)
+        self.assertEqual(last_used_hash, 'success1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_fail_to_log(self, mock_check_call, mock_check_output):
+        mock_check_output.side_effect = subprocess.CalledProcessError(returncode=1,
+                                                                    cmd=['git', '-C', 'archive_path',
+                                                                         'log', '--format=format:%H',
+                                                                         '-n',str(MAX_NUMBER_OF_RECENT_COMMITS)],
+                                                                    stderr=b'Failed to log')
+
+        with self.assertRaises(subprocess.CalledProcessError):
+            fill_message_directory('archive_path', self.tmp_dir, '')
+
+        self.assertEqual(len(os.listdir(self.tmp_dir)), 0)
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_no_hashes_available(self, mock_check_call, mock_check_output):
+        mock_check_output.return_value = b''
+
+        with self.assertRaisesRegex(Exception, 'no commits'):
+            fill_message_directory('', self.tmp_dir, '')
+
+        self.assertEqual(len(os.listdir(self.tmp_dir)), 0)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#still working through a couple of things and haven't tested everything, but this is a rough draft of the changes to come
+
+import subprocess
+import os
+
+def update_archive(archive_path:str):
+    subprocess.check_call(['git', '-C', archive_path,'fetch'])
```
dlatypov: nit: can you add a space between `,'f`

archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#still working through a couple of things and haven't tested everything, but this is a rough draft of the changes to come
+
+import subprocess
+import os
+
+def update_archive(archive_path:str):
```
dlatypov: nit: python style usually says to have a space between names and the annotations, e.g.
```suggestion
def update_archive(archive_path: str):
```

Ditto for the lines below.

archive_updater.py
dlatypov: yes, you'll likely want to mock out the subprocess.check_call() func for testing purposes.
Let's sync offline about this.

archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+    
+def fill_message_directory(archive_path:str, directory_path:str, last_used_commit_hash:str) -> str:
+    process = subprocess.Popen(['git', '-C', archive_path, 'log', '--format=format:%H'], stdout=subprocess.PIPE)
+    output = process.communicate()[0]
+    message_hashes = output.decode('utf-8').split()
+    
+    file_count = 0
+    while(message_hashes[file_count] != last_used_commit_hash):
+        file = os.path.join(directory_path, 'email' + str(file_count) + '.txt')
+        with open(file, 'w+') as output:
```
dlatypov: hmm, do we need `w+`, aka read+write on this file?
I would have thought just `'w'` would have been enough.

archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+import subprocess
+import os
+
+def update_archive(archive_path:str):
+    subprocess.check_call(['git', '-C', archive_path,'fetch'])
+    subprocess.check_call(['git', '-C', archive_path,'pull'])
+    
+def fill_message_directory(archive_path:str, directory_path:str, last_used_commit_hash:str) -> str:
+    process = subprocess.Popen(['git', '-C', archive_path, 'log', '--format=format:%H'], stdout=subprocess.PIPE)
+    output = process.communicate()[0]
```
dlatypov: optional: I personally prefer
```suggestion
    output, _ = process.communicate()
```

archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+
+import subprocess
+import os
+
+def update_archive(archive_path:str):
+    subprocess.check_call(['git', '-C', archive_path,'fetch'])
+    subprocess.check_call(['git', '-C', archive_path,'pull'])
+    
+def fill_message_directory(archive_path:str, directory_path:str, last_used_commit_hash:str) -> str:
+    process = subprocess.Popen(['git', '-C', archive_path, 'log', '--format=format:%H'], stdout=subprocess.PIPE)
```
dlatypov: a comment here about what this function doing would be good.

A docstring about the high-level goal would be good, but an explanation that this is trying to list out every git hash in a comment also would help.
(Along with potentially a TODO to avoid having it show everything every time.
Tangent: not sure what the git history looks like, but we could ask log for all child commits of `last_used_commit_hash`)

archive_updater.py
```diff
@@ -0,0 +1,42 @@
...
+    output = process.communicate()[0]
+    message_hashes = output.decode('utf-8').split()
+    
+    file_count = 0
+    while(message_hashes[file_count] != last_used_commit_hash):
+        file = os.path.join(directory_path, 'email' + str(file_count) + '.txt')
```
dlatypov:
```suggestion
for i, hash in enumerate(message_hashes):
  if hash = last_used_commit_hash:
     break
  file = os.path.join(directory_path, f'email{i}.txt')
```

archive_updater.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2019 Google LLC
```
dlatypov: ah, we might have missed this in previously added files, but you'll want 2020 here.

archive_updater.py
dlatypov: very optional+personal preference:

if we follow, https://google.github.io/styleguide/pyguide.html#22-imports
then we can simplify the names of these funcs a bit, e.g.

```py
  def update(archive_path: str):
    ...
```

then in other files

```py
import archive_updater
...
archive_updater.update(path)
```

Makes it a bit easier to tell where funcs + types are coming from when reading through.

dlatypov (changes requested):
Thanks for pushing this, makes more sense now seeing the code.

As noted below, let's sync offline about how to go about testing this.

archive_updater.py
```diff
@@ -0,0 +1,67 @@
...
+import subprocess
+import os
+
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+
+def update_archive(archive_path:str) -> bool:
+    ''' Fetches updates for the git repo corresponding to the passed in path. 
+    Returns true if fetch is successful, and false otherwise.'''
+    
+    process = subprocess.Popen(['git', '-C', archive_path,'fetch'],
```
dlatypov: We could simplify this func by making it do`subprocess.check_call()` instead.

Can document in the docstring that it throws subprocess.<NameIForgetException> on error.
Or you could catch that and throw a new exception type.
`fill_message_directory` can similarly document that as well so you don't have to return '' to signal error.

gerrit.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
```
dlatypov: we shouldn't be changing the files you didn't add.
2019 is appropriate, since these were added in 2019.

archive_updater_test.py
```diff
@@ -0,0 +1,68 @@
...
+        mock_popen.return_value.communicate.return_value = [b'success1 success2 success3', '']
+        
+        last_used_hash = fill_message_directory('', self.tmp_dir, '')
+        
+        self.assertEqual(len(os.listdir(self.tmp_dir)), 3)
+        for filename in os.listdir(self.tmp_dir):
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertEqual(val, 'SUCCESS')
+        self.assertEqual(last_used_hash,'success1')
```
dlatypov: minor nit: missing spaces after `,` in the `assertEquals()` calls for `last_used_hash` across test cases

archive_updater_test.py
```diff
@@ -0,0 +1,68 @@
...
+        shutil.rmtree(self.tmp_dir)
+        
+    @mock.patch.object(subprocess, 'Popen')
+    @mock.patch.object(subprocess, 'call')
+    def test_fill_message_directory_success(self, mock_call, mock_popen):
+        def writeFile(args, stdout):
+            if 'show' in args:
+                stdout.write('SUCCESS')
+                return
+            stdout.write('FAIL')
```
dlatypov: something we can look at later once the test issue is debugged:

This func doesn't need to actually check the args.
We could do that with [assert_has_calls()](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls) on the mock itself.

As is, should this fail, we'd find out.
But we wouldn't know what the offending args was that caused it to fail.
Would require changing the test code to find that out, as opposed to using assert_has_calls().

dlatypov (changes requested):

archive_updater_test.py
WLIU0628: I'm a little confused as to what this means, do you think I need to change anything right now? So I should use something like assert_has_calls(), but what in the mock am I asserting has been called? I guess I'm just not familiar with it.

dlatypov: Currently we're relying on this function to validate that `git show` is called for each of the files

I.e. if we called something else, the file contents would be different below.

This function doesn't necessarily need to be the place where we do this validation.

```python
  def writeFile(args, stdout):
     stdout.write('called with: ' + ' '.join(args))
   # loop below then checks that 'show' is in each file, perhaps also that it has the right commit, etc.
```

WLIU0628: ah I see okay!

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+        raise Exception(f'Failed to log recent commits, see error: {error.decode()}')
+    message_hashes = output.decode('utf-8').split()
+    
+    if len(message_hashes) == 0:
+        raise Exception('There are no recent commits on the given repo')
+    
+    for i, hash in enumerate(message_hashes):
+        if hash == last_used_commit_hash:
+            break
+        file = os.path.join(directory_path, f'email{i}.txt')
```
WLIU0628: This email{i} naming convention will cause emails to be improperly indexed. First, the lesser the number, the more recent the commit (i.e. the most recent commit will be email0.txt). This can be changed so that it is reverse indexed (i.e. the greater the commit, the more recent) but email11.txt will be indexed before email2.txt. How should I adjust the code for this?

dlatypov: Since email1.txt name already isn't in itself useful (you can find specific messages from it), why not use the SHA?
I.e. `f'email_{hash}.txt`

Those should never change.
But as noted elsewhere, long-term, I don't think we'll be creating files for all of these since the git repo already contains them in a more compact format and we could always run `git show {hash}:m` if we wanted to pull the email contents on demand.

WLIU0628: but the hashes won't necessarily be sorted right? As in a hash corresponding to a more recent commit will not necessarily appear before the hash of a later commit in a alphabetical sort, right? The current indexing system reads the files in alphabetical order (i think) and so files named with alphabetically smaller names will be indexed first, but they might be messages that are more recent and are a reply to a file that has yet to be indexed.

dlatypov: Yes, you're probably getting the files in alphabetical order from `os.listdir()`
But you'd get `email1.text`, `email10.txt`, etc. and then `email2.txt`.

So relying on getting them in chronological order from `os.listdir()` isn't viable once we go past 9 emails (and I hope we do!)

We're going to need the indexer to be able to handle emails out of order.

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+
+import subprocess
+import os
+
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+    
+def fill_message_directory(archive_path: str, directory_path: str, last_used_commit_hash: str) -> str:
+    '''Given a git repo path, directory path, and last used commit hash, this method updates the git repo,
+    then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored in
+    the directory corresponding to the passed in directory path. Throws CalledProcessError and Exception'''
```
dlatypov: this seems like this should become a multiline docstring, esp. given you're calling out "throws CalledProcessError..."

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import os
+
+MAX_NUMBER_OF_RECENT_COMMITS = 2
```
dlatypov, line 19: nit: put a TODO to bump this value up

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+    '''Given a git repo path, directory path, and last used commit hash, this method updates the git repo,
+    then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored in
+    the directory corresponding to the passed in directory path. Throws CalledProcessError and Exception'''
+    
+    fetch = subprocess.check_call(['git', '-C', archive_path, 'fetch'])
+    
+    log = subprocess.Popen(
+        ['git', '-C', archive_path, 'log', '--format=format:%H', '-n',str(MAX_NUMBER_OF_RECENT_COMMITS)],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE)
```
dlatypov: is the reason for the use of Popen() to get stderr and stdout?

See https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError
> **stderr**
> Stderr output of the child process if it was captured by run(). Otherwise, None.

can just use one of the check_* funcs

archive_updater.py
dlatypov: Seems like we've agreed that this loop can process emails out of order.

Can you go and update `file` to not use `i` in the name before we forget about it? :)

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    if len(message_hashes) == 0:
+        raise Exception('There are no recent commits on the given repo')
+    
+    for i, hash in enumerate(message_hashes):
+        if hash == last_used_commit_hash:
+            break
+        file = os.path.join(directory_path, f'email{i}.txt')
+        with open(file, 'w') as f:
+            subprocess.call(['git', '-C', archive_path, 'show', f'{hash}:m'],
```
dlatypov, line 53: nit: do we want to keep track of failures here?

Since I don't think we'll be keeping these files around on disk

So perhaps we can just have the TODO instead, e.g.
`TODO(willliu@google.com): fetch the message contents on demand. We also don't check for errors creating the file`

archive_updater.py
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import os
+
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+
```
dlatypov, line 20: nit: you have some more trailing whitespace in this file.
mind removing them like last time?

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+import subprocess
+import os
+import tempfile
+import shutil
+from archive_updater import fill_message_directory
+from unittest import mock
+
+class ArchiveUpdaterFillMessageDirectoryTest(unittest.TestCase):
+
```
dlatypov: nit: also have trailing whitespace in this file

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+    @mock.patch.object(subprocess, 'call')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_success(self, mock_check_call, mock_call, mock_popen):
+        def writeFile(args, stdout):
+            stdout.write('called with: ' + ' '.join(args))
+        mock_check_call.return_value = ''
+        mock_call.side_effect = writeFile
+        mock_popen.return_value.communicate.return_value = [b'success1 success2 success3', '']
+        
+        last_used_hash = fill_message_directory('abc', self.tmp_dir, '')
```
dlatypov: nit: I'd say `'path'` or `'archive_path'` here

Makes it easier to understand the expectation below, e.g. out of context one might ask:
Why do we want `git show -C abc`?
`git show -C archive_path` is a bit more self evident.

dlatypov (changes requested):

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+# limitations under the License.
+
+import subprocess
+import os
+
+# TODO(willliu@google.com): bump this value up
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+
+def fill_message_directory(archive_path: str, directory_path: str, last_used_commit_hash: str) -> str:
+    '''Given a git repo path, directory path, and last used commit hash, this method updates the git repo,
```
dlatypov: "Given <args>" is a bit redundant with the bit below.

Perhaps something worded like
```python
"""Updates the git repo, and converts up to MAX_NUMBER_OF_RECENT_COMMITS commits into files in..."""
```

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+    then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored in
+    the directory corresponding to the passed in directory path.
+
+    Parameters:
```
dlatypov: Nit, let's call this "Args"

It's a bit shorter and gets the point across.
Also happens to be what google style guide (which you don't have to follow for this) suggests: https://google.github.io/styleguide/pyguide.html#doc-function-raises

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+    archive_path - the path to the email archive
+    directory_path - the directory where files will be temporarily stored
+    last_used_commit_hash - the last commit hash which will be used as the ending point of this function
+
+    Returns the most recent commit hash that was processed in this method call
+
+    Throws CalledProcessError and Exception'''
```
dlatypov: This would be perhaps a bit cleaner following https://google.github.io/styleguide/pyguide.html#doc-function-raises

```python
    Raises:
       CalledProcessError: <explain why/when>
       Exception: <ditto>
```

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+    '''Given a git repo path, directory path, and last used commit hash, this method updates the git repo,
+    then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored in
+    the directory corresponding to the passed in directory path.
+
+    Parameters:
+    archive_path - the path to the email archive
+    directory_path - the directory where files will be temporarily stored
+    last_used_commit_hash - the last commit hash which will be used as the ending point of this function
+
+    Returns the most recent commit hash that was processed in this method call
```
dlatypov: If you want to follow the style guide linked above, this should become something like

```suggestion
    Returns:
        The most recent commit hash processed by this call.
```

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import os
+
+# TODO(willliu@google.com): bump this value up
+MAX_NUMBER_OF_RECENT_COMMITS = 2
+
+def fill_message_directory(archive_path: str, directory_path: str, last_used_commit_hash: str) -> str:
```
dlatypov: nit: `directory_path` is a bit redundant.
How about just `directory`?

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+    '''Given a git repo path, directory path, and last used commit hash, this method updates the git repo,
+    then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored in
+    the directory corresponding to the passed in directory path.
+
+    Parameters:
+    archive_path - the path to the email archive
+    directory_path - the directory where files will be temporarily stored
```
dlatypov: nit: prefer to avoid the passive voice
So
```suggestion
    directory: - where to store ...
```

This feels like relatively common advice, but I've found this link if you want a bit more of an explanation on why to prefer the active voice.
https://developers.google.com/style/voice

This goes for the other comments here as well
"which will be used as the ending point"
"hash that was processed"
etc.

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+    last_used_commit_hash - the last commit hash which will be used as the ending point of this function
+
+    Returns the most recent commit hash that was processed in this method call
+
+    Throws CalledProcessError and Exception'''
+
+    fetch = subprocess.check_call(['git', '-C', archive_path, 'fetch'])
```
dlatypov: check_call doesn't return anything (so an implicit `None` is stored here) and this value isn't used anyways.
```suggestion
    subprocess.check_call(['git', '-C', archive_path, 'fetch'])
```

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+
+    output = subprocess.check_output(
+        ['git', '-C', archive_path, 'log', '--format=format:%H', '-n',str(MAX_NUMBER_OF_RECENT_COMMITS)])
+    message_hashes = output.decode('utf-8').split()
+
+    if len(message_hashes) == 0:
+        raise Exception('There are no recent commits on the given repo')
```
dlatypov: "recent" confuses me a bit.

`git -C <path> log -n 2` will show commits even if they aren't recent.
So we'd only hit this if there are no commits at all in the repo, unless I'm misunderstanding something.

We might want to do
```python
raise Exception('There are no commits in git repo ' + archive_path)
```

archive_updater.py
```diff
@@ -0,0 +1,59 @@
...
+
+    for i, hash in enumerate(message_hashes):
```
dlatypov: We don't need `i` anymore, can do
```suggestion
    for hash in message_hashes:
```

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+                                                 '../lkml-gerrit-bridge/test_data',
+                                                 '5227e860278407173cbfdab594676764153c27e1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'call')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_success(self, mock_check_call, mock_call, mock_check_output):
+        def writeFile(args, stdout):
+            stdout.write('called with: ' + ' '.join(args))
+        mock_check_call.return_value = ''
```
dlatypov: `mock_check_call` shouldn't return anything.
Drop this line.

This also keeps the def of `writeFile` closer to where it's used (the line below this)

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+        for filename in os.listdir(self.tmp_dir):
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertIn('git -C archive_path show', val)
+        self.assertEqual(last_used_hash, 'success1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_fail_to_log(self, mock_check_call, mock_check_output):
+        mock_check_call.return_value = ''
```
dlatypov: ditto, drop this return_value

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+        mock_check_output.side_effect = subprocess.CalledProcessError('Failed to log', 'git log')
+
+        self.assertRaises(subprocess.CalledProcessError, fill_message_directory, '', self.tmp_dir, '')
```
dlatypov: I would prefer

```python
with self.assertRaises(subprocess.CalledProcessError):
  fill_message_directory('', self.tmp_dir, '')
```

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertIn('git -C archive_path show', val)
+        self.assertEqual(last_used_hash, 'success1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_fail_to_log(self, mock_check_call, mock_check_output):
+        mock_check_call.return_value = ''
+        mock_check_output.side_effect = subprocess.CalledProcessError('Failed to log', 'git log')
```
dlatypov: I don't think this is initializing the exception like you intended.
You're setting the `returncode` to 'Failed to log' in this case.

Do you want to use named parameters here to make it more clear?

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+        self.assertRaises(subprocess.CalledProcessError, fill_message_directory, '', self.tmp_dir, '')
+
+        self.assertEqual(len(os.listdir(self.tmp_dir)), 0)
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_no_hashes_available(self, mock_check_call, mock_check_output):
+        mock_check_call.return_value = ''
```
dlatypov: ditto, drop this return_value

archive_updater_test.py
```diff
@@ -0,0 +1,65 @@
...
+        mock_check_output.return_value = b''
+
+        self.assertRaises(Exception, fill_message_directory, '', self.tmp_dir, '')
```
dlatypov:
```suggestion
        with self.assertRaisesRegex(Exception, 'no commits'):
            fill_message_directory('', self.tmp_dir, '')
```

dlatypov (changes requested):

archive_updater_test.py
```diff
@@ -0,0 +1,64 @@
...
+        for filename in os.listdir(self.tmp_dir):
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertIn('git -C archive_path show', val)
+        self.assertEqual(last_used_hash, 'success1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_fail_to_log(self, mock_check_call, mock_check_output):
+        mock_check_output.side_effect = subprocess.CalledProcessError(returncode = 'Failed to log',
```
dlatypov: nit: `returncode` is supposed to be an int.
Can you change this to something like

`returncode=1, cmd=['git', ...], stderr=b'Failed to log'`

dlatypov (approved):
One minor nit, but otherwise LGTM.

bjh83 (approved):

archive_updater_test.py
```diff
@@ -0,0 +1,68 @@
...
+            with open(os.path.join(self.tmp_dir,filename), "r") as file:
+                val = file.read()
+                self.assertIn('git -C archive_path show', val)
+        self.assertEqual(last_used_hash, 'success1')
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(subprocess, 'check_call')
+    def test_fill_message_directory_fail_to_log(self, mock_check_call, mock_check_output):
+        mock_check_output.side_effect = subprocess.CalledProcessError(returncode=1,
+                                                                    cmd=['git', '-C', 'archive_path',
```
dlatypov, line 46: note: didn't need to specify the full cmd here since it's not relevant to our code or the test, but it's fine either way.

dlatypov (approved):

PR #5, opened 2020-09-29
Improving conversion logic

WLIU0628:
Created a DAO class which wraps a hashmap to simplify transitioning to postgres later. Also, changed index updating so that it returns all new messages so that they may be properly processed across batches.

### DIFF
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -16,30 +16,36 @@ import email
 import os
 from typing import List, Dict
 from setup_gmail import Message
+from message_dao import MessageDao
 
 class ArchiveMessageIndex(object):
-    def __init__(self):
-        # Maps message subject (str) to starting message of email thread
-        self._key_to_thread_start : Dict[str, Message] = {}
-        # Maps message.id to message
-        self._messages_seen : Dict[str, Message] = {}
+    def __init__(self, message_dao : MessageDao):
+        self._message_dao = message_dao
 
-    def find_thread(self, key: str) -> Message:
-        return self._key_to_thread_start[key]
+    #TODO(willliu@google.com): Process the new messages to determine which ones need to be uploaded
+    def update(self, data_dir: str) -> List[Message]:
+        """ Updates index with messages in the passed in directory.
+        Returns a list of new messages """
 
-    def update(self, data_dir: str):
         new_messages : List[Message] = []
 
         for filename in os.listdir(data_dir):
             if not filename.endswith(".txt"):
                 continue
             email = generate_email_from_file(os.path.join(data_dir, filename))
-            if email.id not in self._messages_seen:
+            if not self._message_dao.get(email.id):
                 new_messages.append(email)
         self._add_messages_to_index(new_messages)
+        return new_messages
 
     def size(self):
-        return len(self._messages_seen)
+        return self._message_dao.size()
+
+    def find(self, message_id : str) -> Message:
+        message = self._message_dao.get(message_id)
+        if message is None:
+            raise ValueError(f'Could not find message: {message_id}')
+        return message
 
     def _add_messages_to_index(self, new_messages : List[Message]):
         """ Iterates through all new emails and links together emails that form a thread by populating message.children. """
@@ -48,31 +54,21 @@ class ArchiveMessageIndex(object):
 
         # First iterates through all messages to distinguish between 1.) replies and 2.) the start of a thread.
         for message in new_messages:
-            if message.id not in self._messages_seen:
-                self._messages_seen[message.id] = message
+            if not self._message_dao.get(message.id):
+                self._message_dao.store(message)
 
             # If message is a reply, it needs to be associated with an existing thread
             if message.in_reply_to is not None:
                 need_parent.append(message)
                 continue
 
-            # If message doesn't have the LKML style [PATCH v4 00/11],
-            # we don't need to put it in the index.
-            key = message.get_key()
-            if key:
-                self._key_to_thread_start[key] = message
-
         for message in need_parent:
-            if message.in_reply_to not in self._messages_seen:
+            parent = self._message_dao.get(message.in_reply_to)
+            if not parent:
                 print("Could not find parent email, dropping " + message.subject)
                 continue
-            parent = self._messages_seen[message.in_reply_to]
             parent.children.append(message)
-
-def find_thread(key: str) -> Message:
-    archive_index = ArchiveMessageIndex()
-    archive_index.update('test_data')
-    return archive_index.find_thread(key)
+            self._message_dao.store(parent)
 
 def generate_email_from_file(file: str):
     with open(file, "r") as raw_email:
--- a/patch_parser_test.py
+++ b/patch_parser_test.py
@@ -1,11 +1,16 @@
 import unittest
 from patch_parser import parse_comments
-from archive_converter import find_thread, generate_email_from_file
+from archive_converter import ArchiveMessageIndex, generate_email_from_file
+from message_dao import MessageDao
 
 class PatchParserTest(unittest.TestCase):
 
     def test_parse_comments_for_single_email_thread(self):
-        patchset = parse_comments(find_thread('PATCH v2 1/2'))
+        archive_index = ArchiveMessageIndex(MessageDao())
+        archive_index.update('test_data')
+        patchset = parse_comments(
+            archive_index.find(
+                '<20200827144112.v2.1.I6981f9a9f0c12e60f8038f3b574184f8ffc1b9b5@changeid>'))
 
         self.assertTrue(len(patchset.patches) > 0)
         first_patch = patchset.patches[0]
@@ -15,7 +20,9 @@ class PatchParserTest(unittest.TestCase):
         self.assertEqual(first_patch.comments, [])
 
     def test_parse_comments_for_multi_email_thread_with_cover_letter(self):
-        patchset = parse_comments(find_thread('PATCH v2 0/4'))
+        archive_index = ArchiveMessageIndex(MessageDao())
+        archive_index.update('test_data')
+        patchset = parse_comments(archive_index.find('<20200831110450.30188-1-boyan.karatotev@arm.com>'))
 
         self.assertEqual(len(patchset.patches), 4)
         first_patch = patchset.patches[0]
--- a/archive_converter_test.py
+++ b/archive_converter_test.py
@@ -1,5 +1,8 @@
 import unittest
 from archive_converter import generate_email_from_file, ArchiveMessageIndex
+from message_dao import MessageDao
+from typing import List
+from setup_gmail import Message
 
 class ArchiveConverterTest(unittest.TestCase):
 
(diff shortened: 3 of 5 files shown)


archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
-    return archive_index.find_thread(key)
+            
+            # determines whether the parent of the given email is part of this update or not
+            if parent not in parent_messages:
+                if re.match(r'\[.+ (\d+)/(\d+)\] .+', message.subject):
+                    patches_to_upload.append(message)
+                else:
+                    patches_with_new_comments.add(parent)
+                    
+        return tuple(patches_to_upload), tuple(patches_with_new_comments)
```
dlatypov: these should probably be lists and not tuples if you're returning a variable sized collection of messages

archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
-        self._messages_seen : Dict[str, Message] = {}
-
-    def find_thread(self, key: str) -> Message:
-        return self._key_to_thread_start[key]
-
-    def update(self, data_dir: str):
+    def __init__(self, message_dao : MessageDao):
+        self._message_dao = message_dao
+        
+    def update(self, data_dir: str) -> Tuple[Tuple[Message, ...], Tuple[Message, ...]]:
```
dlatypov: this should probably now have a docstring given it has a not-too-obvious return type

patch_parser_test.py
```diff
@@ -1,11 +1,16 @@
...
 from patch_parser import parse_comments
-from archive_converter import find_thread, generate_email_from_file
+from archive_converter import ArchiveMessageIndex, generate_email_from_file
+from message_dao import MessageDao
 
 class PatchParserTest(unittest.TestCase):
 
     def test_parse_comments_for_single_email_thread(self):
-        patchset = parse_comments(find_thread('PATCH v2 1/2'))
+        archive_index = ArchiveMessageIndex(MessageDao())
```
dlatypov, line 9: should we have at least one test case verify the return value of `update()`?

archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
+        if message is None:
+            raise ValueError(f'Could not find message: {message_id}')
+        return message
 
-    def _add_messages_to_index(self, new_messages : List[Message]):
+    def _add_messages_to_index(self, new_messages : List[Message]) -> Tuple[Tuple[Message, ...], Tuple[Message, ...]]:
         """ Iterates through all new emails and links together emails that form a thread by populating message.children. """
 
         need_parent : List[Message] = []
+        parent_messages : Set[Message] = set([])
```
```suggestion
        parent_messages : Set[Message] = set()
```

archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
+        patches_to_upload : List[Message] = []
+        patches_with_new_comments : Set[Message] = set([])
```
dlatypov: ditto of above

patch_parser_test.py
```diff
@@ -1,11 +1,16 @@
...
+        archive_index.update('test_data')
+        patchset = parse_comments(
+            archive_index.find(
+                '<20200827144112.v2.1.I6981f9a9f0c12e60f8038f3b574184f8ffc1b9b5@changeid>'))
```
dlatypov, line 13: can we include an assert on the subject or similar?
Would make the test a bit easier to read, imo.

This seems to be patch6.txt, aka
> Subject: [PATCH v2 1/2] Input: i8042 - Prevent intermixing i8042 commands

archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
             parent.children.append(message)
-
-def find_thread(key: str) -> Message:
-    archive_index = ArchiveMessageIndex()
-    archive_index.update('test_data')
-    return archive_index.find_thread(key)
+            
+            # determines whether the parent of the given email is part of this update or not
+            if parent not in parent_messages:
+                if re.match(r'\[.+ (\d+)/(\d+)\] .+', message.subject):
```
dlatypov: this feels like it should be a helper function on Message.

Consider adding such a function in a parent commit.
e.g. 
```python
  def is_patch() -> bool:
     ...
  def patch_index() -> Tuple[int, int]:
    ... # can move parse_set_index from patch_parser.py here too
 
```

archive_converter.py
```diff
@@ -14,65 +14,71 @@
...
 
         for filename in os.listdir(data_dir):
             if not filename.endswith(".txt"):
                 continue
             email = generate_email_from_file(os.path.join(data_dir, filename))
-            if email.id not in self._messages_seen:
+            if not self._message_dao.get(email.id):
                 new_messages.append(email)
-        self._add_messages_to_index(new_messages)
+        return self._add_messages_to_index(new_messages)
```
dlatypov: Is there a reason we can't just return `new_messages` here?
Do we need to distinguish between the two different types of new messages?

If so, let's consider augmenting the commit message with a description and/or a code comment.

dlatypov (changes requested):
Hmm, the commit message is a bit vague.
We haven't been paying too much attention to the messages in the past, but could you
* clarify how it is being "adjusted?"
* what "process files async" means?

Optional: also split out the Dao refactor into a parent commit and explain in the commit message there a bit about the rationale (i.e. that the change doesn't have behavior now but is intended to let us use Postgres in the future)

archive_converter.py
WLIU0628: can you return a tuple of lists?

archive_converter.py
WLIU0628: > Optional: also split out the Dao refactor into a parent commit and explain in the commit message there a bit about the rationale (i.e. that the change doesn't have behavior now but is intended to let us use Postgres in the future)

How would I go about doing this?

patch_parser_test.py
WLIU0628: do you think I should hard code in the order? Like, I can expect patch6 to be the first value, and thread_patch0 to be the second.

archive_converter.py
WLIU0628: I can't seem to find resources online to add things in a parent commit. would I basically be rebasing?

archive_converter.py
WLIU0628: so I think it would be fine to return new_messages, but that just leaves the processing into two different types of new_messages for later? I think we need two different types of new messages because not all new messages are directly uploaded. My current understanding (which may be wrong) is that comments are added as children to parent patches and then the upload_all_comments function is called. Also, patchsets are all uploaded together so some processing needs to be done to see whether the start of the patchset is also within the current batch of emails. If it is, then only the start of the patchset needs to be uploaded (as it will upload the rest as well). If it is not, then the reply must be uploaded individually because the start of the patchset has already been uploaded and should not be uploaded twice.

archive_converter.py
WLIU0628: idk why this showed up here, but I meant to ask this in reply to the review message.

patch_parser_test.py
WLIU0628: i'm not too sure what this means. I assert on the subject later in that test, should I be asserting on the find? like self.assertEquals(archive_index.find('hash').subject, "Subject: [PATCH v2 1/2] Input: i8042 - Prevent intermixing i8042 commands"))

archive_converter.py
bjh83: ? Are you asking how to edit a parent commit in a chain of commits? If so, you will probably want to do an interactive rebase ("git rebase -i <parent-commit>^"); you can read more about it here: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#_changing_multiple

patch_parser_test.py
dlatypov: I just meant adding something like

```python
self.assertEqual(first_patch.subject, 'our subject')
```

The goal is to just make it more clear to the test reader what `first_patch` is expected to be.
A patch with no comments on it isn't much to go on.
It'd be easier to go grep for a few words from the subject then to copy-paste the message id (e.g. it might be unclear if the message id will appear exactly in the same form in the testdata files)

archive_converter.py
dlatypov: Meta: adding some '\n' in comments makes it a bit easier to read :)

Yes, we need to treat patches different from comments.
We'd want to upload all patches first and then add comments in the right places.

But we can distinguish between the types of messages later.
As suggested below, the check you're doing now could be moved into the message class.

We could then partition the messages into the two types afterwards fairly easily with that helper func.
So this function doesn't seem like it needs to take on this additional responsibility.

General tip: The more things a function/class is responsible, the harder it is to understand.

patch_parser_test.py
WLIU0628: I thought the patches didn't have a subject field? Once they transform from message to patches, I don't think they have the same fields. I try to do the subject comparison by doing the self.assertIn on line 19. Should I comment that it's the subject comparison?

archive_converter.py
WLIU0628: so for now, would you say it's best to just write a todo and have it return new_messages?

dlatypov: SGTM.

patch_parser_test.py
dlatypov: Ah yes, you're right.
I didn't notice that GitHub hid 3 lines from me which had the `self.assertIn` I was asking you to write.

(Why would it collapse just those few lines? 😞 )

bjh83 (approved):

archive_converter.py
```diff
@@ -48,32 +54,21 @@ def _add_messages_to_index(self, new_messages : List[Message]):
...
 
-            # If message doesn't have the LKML style [PATCH v4 00/11],
-            # we don't need to put it in the index.
-            key = message.get_key()
-            if key:
-                self._key_to_thread_start[key] = message
-
         for message in need_parent:
-            if message.in_reply_to not in self._messages_seen:
+            if not self._message_dao.get(message.in_reply_to):
```
dlatypov: Since `get()` will eventually query the db, should avoid doing `get()` twice here?

I.e.

```python
parent = self._message_dao.get(message.in_reply_to)
if not parent:
  ...
parent.children.append(message)
```

archive_converter.py
```diff
@@ -48,32 +54,21 @@ def _add_messages_to_index(self, new_messages : List[Message]):
...
                 print("Could not find parent email, dropping " + message.subject)
                 continue
-            parent = self._messages_seen[message.in_reply_to]
+            parent = self.find(message.in_reply_to)
```
dlatypov: is finding `parent` every time from the DAO actually correct?

For the in-memory impl. this will work.
But if this is a read from a DB, `parent.children` won't reflect previous calls to `append` below.

Are we planning on storing the children message ids in the DB?
Do we then have `self._message_dao.store(parent)` after we `append()`?

dlatypov (changes requested):
Overall LGTM.
We might need to tweak how we use the DAO to make it more DB-friendly.

Otherwise we might forget and run into bugs when trying to introduce the real DAO.

archive_converter.py
WLIU0628: this is something I completely forgot about, thank you!

dlatypov (approved):
LGTM, thanks!

PR #6, opened 2020-10-07
Added server loop implementation

WLIU0628:
Server loop now handles determining when to set up the index (if DAO is unable to find data/Index is empty), setting up the index (using a epoch hash to act as the 'last relevant commit' and building the index with every message since then), and then running an update, convert, upload, clear loop as the main functionality of the server.


setup_gmail.py
```diff
@@ -61,6 +61,11 @@ def get_key(self):
             return match.group(1)
         else:
             return None
+        
+    def is_patch(self) -> bool:
```
dlatypov, line 66: in this commit, do want to update patch_parser.py to use this?

I.e. in filter_patches_and_cover_letter_replies()

main.py
```diff
@@ -0,0 +1,146 @@
...
+EPOCH_HASH = 'cc49e216e3fdff0ffed7675dc7215aba5e3d05cc'
+GIT_PATH = '../linux-kselftest/git/0.git'
+FILE_DIR = 'index_files'
+
+def remove_files(file_dir : str):
+    files = glob.glob(f'{file_dir}/*')
+    for f in files:
+        os.remove(f)
+
+# returns parents, then replies as lists
```
dlatypov: nit: turn this comment into a docstring

main.py
```diff
@@ -0,0 +1,146 @@
...
+                parents.append(message)
+                continue
+            replies.append(message)
+    return (parents, replies)
+    
+
+def main():
+    gerrit_url = 'https://linux-review.googlesource.com'
+    gob_url = 'http://linux.googlesource.com'
+    rest = get_gerrit_rest_api('gerritcookies', gerrit_url)
```
dlatypov: optional: (esp. since this is how the code is in gerrit.py)
this var isn't used elsewhere.
Can we avoid declaring it?

Perhaps
```python
gerrit = Gerrit(get_gerrit_rest_api(...))
```

or add a parent commit so we can do
```python
gerrit = NewGerrit(cookie_jar_path='gerritcookies', url=gerrit_url)
```

main.py
```diff
@@ -0,0 +1,146 @@
...
+    gerrit = Gerrit(rest)
+    gerrit_git = GerritGit(git_dir='gerrit_git_dir',
+                           cookie_jar_path='gerritcookies',
+                           url=gob_url, project='linux/kernel/git/torvalds/linux', branch='master')
+    message_dao = MessageDao()
+    archive_index = ArchiveMessageIndex(message_dao)
+    last_hash : str
+    # Check if database is populated
+    if(archive_index.size() == 0):
```
dlatypov: Perhaps I'm overlooking something, but I think we can avoid this big `if` that has a good amount of overlap with the main loop.

Could we refactor this code so the fill+update+clear steps only appear in the main loop?

main.py
```diff
@@ -0,0 +1,146 @@
...
+            # Reply is a comment that's parent is not in this batch of messages. It's parent's comments should be reuploaded
+            else:
+                messages_with_new_comments.add(message.in_reply_to)
+                
+        # Start uploading files
+        for message_id in messages_to_upload:
+            try:
+                email_thread = archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                gerrit_git.apply_patchset_and_cleanup(patchset)
```
dlatypov: this all seems like it should work.

One thing to note is that the apply_patchset_and_cleanup() func will create, clone, and delete the gir dir for every single call.

We might want to refactor git.py a bit before this change.

dlatypov (changes requested):
Re commit description: it could be useful to use bullet points.

E.g.
```commit
Server loop handles
* setting up the index ...
   * <more details>
* running update, ....
```

It would be good to call out the `git log` change more clearly than an aside inside of some parens.

As noted in the file comments, I think some initial cleanup of the other files here might be a good idea.

main.py
```diff
@@ -0,0 +1,146 @@
...
+                pass
+            else:
+                print(f'Failed to find {message.id} in index.')
+                pass
+        remove_files(FILE_DIR)
+    else:
+        last_hash = message_dao.get_lash_hash()
+    
+    # Start the main server loop
+    while(True):
```
bjh83: Nit: parens are unpythonic.

main.py
```diff
@@ -0,0 +1,146 @@
...
+            except ValueError:
+                print(f'Failed to find {message_id} in index.')
+                pass
+            else:
+                print(f'Failed to find {message_id} in index.')
+                pass
+        
+        # Clean up directory for next loop
+        remove_files(FILE_DIR)
+        time.sleep(10)
```
bjh83: This method is  fairly huge. Can we maybe split it up into some logical helper functions?

main.py
```diff
@@ -0,0 +1,146 @@
...
+        
+        
+    
+    
+
```
bjh83, line 142: Nit: what's with all the spaces. Isn't 2 normal?

main.py
```diff
@@ -0,0 +1,146 @@
...
+        lash_hash = fill_message_directory(GIT_PATH, FILE_DIR, last_hash)
+        new_messages = archive_index.update(FILE_DIR)
+        
+        # Differentiate between messages to upload and comments
+        messages_to_upload : List[str] = []
+        messages_with_new_comments : Set[str] = set()
+        parent_patches : Set[str] = set()
+        parents, replies = split_parent_and_reply_messages(new_messages)
+        
+        # First separate between parents and replies. All parents of patchsets will be uploaded
```
bjh83, line 81: I think this comment should go above the preceding line (the newline too).

main.py
bjh83: Yeah, this is going to be pretty slow. I don't care if we do it before this change or after, but that refactor definitely needs to happen.

main.py
bjh83: Sorry, to be clear, I mean parens in a while loop like that.

main.py
```diff
@@ -0,0 +1,146 @@
...
+                pass
+            else:
+                print(f'Failed to find {message_id} in index.')
+                pass
+            
+        for message_id in messages_with_new_comments:
+            try:
+                email_thread = archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                upload_all_comments(gerrit, patchset)
```
dlatypov: actually, this won't work as-is.

Patch.change_id is only set by `push_patch`
We're missing a change to allow the DAO to persist the mapping of patches to change ids

dlatypov: Forgot to say: ... and `upload_all_comments()` is just a loop around `upload_comments_for_patch()` which assumes `patch.change_id` is set.

Otherwise it can't tell where to push the comments to.

main.py
WLIU0628: I'll do the first option for now!

main.py
```diff
@@ -0,0 +1,140 @@
...
+        parents : List[Message] = []
+        replies : List[Message] = []
+        for message in messages:
+                if not message.in_reply_to:
+                    parents.append(message)
+                    continue
+                replies.append(message)
+        return (parents, replies)
+    
+    def update_message_dir(self) -> List[Message]:
```
dlatypov: note: unlike in C, these functions don't have to be defined in the file before where they are used.

I personally would find this class a bit clearer to read if we instead ordered the funcs like

```python
# maybe can leave the static methods at top
def run(self):
   ...

def update_convert_upload(self):
  ...

# Then we can organize the rest of the funcs in the order they run
def update_message_dir(self):
   ...
```

main.py
```diff
@@ -0,0 +1,140 @@
...
+    @staticmethod
+    def remove_files(file_dir : str):
+        files = glob.glob(f'{file_dir}/*')
+        for f in files:
+            os.remove(f)
+    
+    @staticmethod
+    def split_parent_and_reply_messages(messages : List[Message]) -> Tuple[List[Message], List[Message]]:
+        ''' Splits a list of messages into parent (first email in a thread) and replies.
+        Returns a list of parents, then replies '''
```
dlatypov: nit: I think we can drop this second docstring line.

main.py
```diff
@@ -0,0 +1,140 @@
...
+                email_thread = self.archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                self.gerrit_git.apply_patchset_and_cleanup(patchset, self.message_dao)
+                find_and_label_all_revision_ids(self.gerrit, patchset)
+                upload_all_comments(self.gerrit, patchset)
+            except ValueError:
+                print(f'Failed to find {message_id} in index.')
+                pass
+            except:
+                print('Failed because of an unexpected error.')
```
dlatypov: Should we show this unexpected error?
Or should we reraise the error or just not catch the error in the first place?

main.py
```diff
@@ -0,0 +1,140 @@
...
+            try:
+                email_thread = self.archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                upload_all_comments(self.gerrit, patchset)
+            except ValueError:
+                print(f'Failed to find {message_id} in index.')
+                pass
+            except:
+                print('Failed because of an unexpected error. Comments')
+                pass
```
dlatypov: ditto the comments above.

main.py
```diff
@@ -0,0 +1,140 @@
...
+    
+    def upload_messages(self, messages_to_upload : List[str]):
+        for message_id in messages_to_upload:
+            try:
+                email_thread = self.archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                self.gerrit_git.apply_patchset_and_cleanup(patchset, self.message_dao)
+                find_and_label_all_revision_ids(self.gerrit, patchset)
+                upload_all_comments(self.gerrit, patchset)
+            except ValueError:
```
dlatypov: optional: we should probably introduce a more specific error type here instead of overloading a bulitin one like `ValueError`

Doesn't have to happen in this PR necessarily.
Can consider adding a todo
```python
except ValueError:  # TODO(willliu@google.com): use a more specific exception
```

main.py
```diff
@@ -0,0 +1,140 @@
...
+        
+        # Determine which of the replies should be uploaded
+        for message in replies:
+            if message.in_reply_to in parent_patches:
+                continue
+            
+            # Reply is a patch to be uploaded (as the parent of patchset is not in new_messages)
+            if message.is_patch():
+                messages_to_upload.append(message.id)
+            # Reply is a comment that's parent is not in this batch of messages. It's parent's comments should be reuploaded
```
dlatypov: nit: its

main.py
```diff
@@ -0,0 +1,140 @@
...
+            else:
+                messages_with_new_comments.add(message.in_reply_to)
+        
+        self.upload_messages(messages_to_upload)
```
dlatypov, line 103: optional: could be useful to log here something like
> Uploading 42 new changes...
> Uploading 36 new comments...

main.py
```diff
@@ -0,0 +1,140 @@
...
+from archive_converter import ArchiveMessageIndex
+from archive_updater import fill_message_directory
+from gerrit import upload_all_comments, get_gerrit_rest_api, Gerrit, find_and_label_all_revision_ids
+from git import GerritGit
+from message_dao import MessageDao
+from patch_parser import map_comments_to_gerrit, parse_comments, Patch, Patchset
+from setup_gmail import Message
+from typing import List, Set, Tuple
+
+GIT_PATH = '../linux-kselftest/git/0.git'
```
dlatypov, line 30: nit: put a TODO here.
We don't currently document the requirement that a user runs

```sh
git clone --mirror https://lore.kernel.org/linux-kselftest/0 linux-kselftest/git/0.git
```
in a directory adjacent to this repo

So long term this should probably be a flag/argument to main.

main.py
```diff
@@ -0,0 +1,140 @@
...
+FILE_DIR = 'index_files'
```
dlatypov, line 31: fyi, the code currently errors out if this dir doesn't exist beforehand.

Might be useful to include a step somewhere to create it if it doesn't exist.
```python
os.makedirs(FILE_DIR, exist_ok=True)
```

main.py
```diff
@@ -0,0 +1,140 @@
...
+        for message_id in messages_to_upload:
+            try:
+                email_thread = self.archive_index.find(message_id)
+                patchset = parse_comments(email_thread)
+                self.gerrit_git.apply_patchset_and_cleanup(patchset, self.message_dao)
+                find_and_label_all_revision_ids(self.gerrit, patchset)
+                upload_all_comments(self.gerrit, patchset)
+            except ValueError:
+                print(f'Failed to find {message_id} in index.')
+                pass
```
dlatypov: nit: pass is not needed here.
If you want to be more explicit that the loop continues, perhaps use `continue` here?

dlatypov (commented):
Just some general most stylistic comments.
Feel free to address when you have time.

I haven't found anything to indicate what might be causing issues with not uploading patches beyonds the fact we don't check for errors from our git commands in git.py.
My working theory is that silently something failed there and caused issues that bled over to other patches (we know at least that a `git am` failed)

git.py
```diff
@@ -85,10 +86,17 @@ def shallow_clone(self, depth=1):
...
-
+        
     def apply_patch(self, patch: Patch):
         print(patch.text_with_headers)
-        return self._git.am(patch.text_with_headers)
+        try:
+            output = self._git.am(patch.text_with_headers)
+            return output
+        except subprocess.CalledProcessError:
+            print('Failed to amend. Aborting...')
```
dlatypov: nit: s/ammend/apply patch.
> git-am - Apply a series of patches from a mailbox

Reminder you still need to make the change for `git()` to error out when the cmd fails via `check_call` or similar.

WLIU0628: does check = true in the run part cover that?

dlatypov: Yes, `check=True` should be sufficient to cause it to raise the Exception we expect.

dlatypov (approved):

PR #8, opened 2020-10-21
Add logging

WLIU0628:
Added abseil logging to de-clutter terminal output. To output logged errors to terminal, run: python3 main.py --stderrthreshold=error

### DIFF
--- a/archive_converter.py
+++ b/archive_converter.py
@@ -14,7 +14,8 @@
 
 import email
 import os
-import subprocess
+
+from absl import logging
 
 from typing import List, Dict, Optional
 from setup_gmail import Message
@@ -24,7 +25,6 @@ class ArchiveMessageIndex(object):
     def __init__(self, message_dao : MessageDao):
         self._message_dao = message_dao
 
-    #TODO(willliu@google.com): Process the new messages to determine which ones need to be uploaded
     def update(self, data_dir: str) -> List[Message]:
         """ Updates index with messages in the passed in directory.
         Returns a list of new messages """
@@ -67,7 +67,7 @@ class ArchiveMessageIndex(object):
         for message in need_parent:
             parent = self._message_dao.get(message.in_reply_to)
             if not parent:
-                print("Could not find parent email, dropping " + message.subject)
+                logging.info('Could not find parent email, dropping %s', message.debug_info())
                 continue
             parent.children.append(message)
             self._message_dao.store(parent)
@@ -76,13 +76,12 @@ def generate_email_from_file(file: str) -> Optional[Message]:
     with open(file, "r") as raw_email:
         try:
             compiled_email = email.message_from_string(raw_email.read())
-            return _email_to_message(compiled_email)
+            return _email_to_message(compiled_email, file[12:-4])
         except Exception as e:
-            # TODO: Log exception
-            print(e)
+            logging.error('Failed to generate email from archive. Error: %s', e)
             return None
 
-def _email_to_message(compiled_email) -> Message:
+def _email_to_message(compiled_email, archive_hash) -> Message:
     content = []
     if compiled_email.is_multipart():
         for payload in compiled_email.get_payload():
@@ -93,4 +92,5 @@ def _email_to_message(compiled_email) -> Message:
                    compiled_email['subject'],
                    compiled_email['from'],
                    compiled_email['In-Reply-To'],
-                   content)
+                   content,
+                   archive_hash)
--- a/archive_updater.py
+++ b/archive_updater.py
@@ -15,6 +15,8 @@
 import subprocess
 import os
 
+from absl import logging
+
 def fill_message_directory(archive_path: str, directory: str, last_used_commit_hash: str) -> str:
     '''Updates the git repo, then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored
     in the directory corresponding to the passed in directory path.
@@ -39,7 +41,7 @@ def fill_message_directory(archive_path: str, directory: str, last_used_commit_h
     message_hashes = output.decode('utf-8').split()
 
     if len(message_hashes) == 0:
-        print(f'There are no commits in git repo: {archive_path}')
+        logging.warning('There are no commits in git repo: %s', archive_path)
         return last_used_commit_hash
 
     for hash in message_hashes:
--- a/gerrit.py
+++ b/gerrit.py
@@ -18,6 +18,7 @@ import base64
 import pickle
 import os.path
 import re
+from absl import logging
 from git import GerritGit
 from googleapiclient.discovery import build
 from google_auth_oauthlib.flow import InstalledAppFlow
@@ -76,9 +77,9 @@ class Gerrit(object):
 
 def find_and_label_revision_id(gerrit: Gerrit, patch: Patch):
     change_info = gerrit.get_change(patch.change_id)
-    print(change_info)
+    logging.info('Change info: %s', change_info)
     revision_id = change_info['current_revision']
-    print(revision_id)
+    logging.info('Revision ID: %s', revision_id)
     patch.revision_id = revision_id
 
 def find_and_label_all_revision_ids(gerrit: Gerrit, patchset: Patchset):
@@ -107,8 +108,8 @@ def upload_comments_for_patch(gerrit: Gerrit, patch: Patch):
             'labels': {'Code-Review': 0},
             'comments': file_comments
     }
-    print('review = ' + str(review))
-    print('set_review response = ' + str(gerrit.set_review(change_id=patch.change_id, revision_id=patch.revision_id, review=review)))
+    logging.info('review = %s', review)
+    logging.info('set_review response = %s', gerrit.set_review(change_id=patch.change_id, revision_id=patch.revision_id, review=review))
 
 def upload_all_comments(gerrit: Gerrit, patchset: Patchset):
     map_comments_to_gerrit(patchset)
--- a/git.py
+++ b/git.py
@@ -18,11 +18,10 @@ import shutil
 import subprocess
 import message_dao
 from patch_parser import Patch, Patchset
+from absl import logging
 
 def git(verb: str, *args, cwd=None, input=None) -> str:
-    print(input)
-    print(f'git {verb} ' + ' '.join(args))
-    print(list(args))
+    logging.debug('Running\ngit %s %s\n with input: %s', verb, ' '.join(args), input)
     result = subprocess.run(['git', verb] + list(args),
                             cwd=cwd, input=input,
                             text=True,
@@ -30,7 +29,7 @@ def git(verb: str, *args, cwd=None, input=None) -> str:
                             stderr=subprocess.STDOUT,
                             stdout=subprocess.PIPE)
     stdout = str(result.stdout)
-    print('stdout: ' + stdout)
+    logging.info('git %s stdout: %s', verb, stdout)
     return stdout
 
 class Git(object):
@@ -59,18 +58,18 @@ flags=re.MULTILINE)
 GERRIT_CHANGE_ID_MATCHER = re.compile(r'^https://[\w/+.-]+\+/(\d+)$')
 
 def _parse_gerrit_patch_push(gerrit_result: str) -> str:
-    print(gerrit_result)
+    logging.info('%s', gerrit_result)
     match = GERRIT_PUSH_MATCHER.search(gerrit_result)
     if match is None:
       raise ValueError(f'Could not find change url from gerrit output: {gerrit_result}')
     change_url = match.group(1)
-    print('change_url = ' + change_url)
+    logging.info('change_url = %s', change_url)
 
     match = GERRIT_CHANGE_ID_MATCHER.match(change_url)
     if match is None:
       raise ValueError(f'Could not extract change id from gerrit output: {gerrit_result}')
     change_id = match.group(1)
-    print('change_id = ' + change_id)
+    logging.info('change_id = %s', change_id)
     return change_id
 
 CURL_CHANGE_ID_CMD = 'curl -Lo `git rev-parse --git-dir`/hooks/commit-msg https://gerrit-review.googlesource.com/tools/hooks/commit-msg ; chmod +x `git rev-parse --git-dir`/hooks/commit-msg'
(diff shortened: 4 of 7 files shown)


archive_updater.py
```diff
@@ -15,44 +15,46 @@
...
-
+    
     output = subprocess.check_output(
         ['git', '-C', archive_path, 'log', f'{last_used_commit_hash}..', '--format=format:%H'])
     message_hashes = output.decode('utf-8').split()
-
+    
     if len(message_hashes) == 0:
-        print(f'There are no commits in git repo: {archive_path}')
+        logging.info(f'There are no commits in git repo: {archive_path}')
```
dlatypov: nit: this feels a bit more like a warning

gerrit.py
```diff
@@ -107,8 +108,8 @@ def upload_comments_for_patch(gerrit: Gerrit, patch: Patch):
             'labels': {'Code-Review': 0},
             'comments': file_comments
     }
-    print('review = ' + str(review))
-    print('set_review response = ' + str(gerrit.set_review(change_id=patch.change_id, revision_id=patch.revision_id, review=review)))
+    logging.info('review = %s', str(review))
```
dlatypov: %s should mean you don't need to cast to str anymore

```
>>> from absl import logging
>>> logging.error('dict: %s', {'a': 1, 42: 'b'})
ERROR:absl:dict: {'a': 1, 42: 'b'}
```

gerrit.py
```diff
@@ -107,8 +108,8 @@ def upload_comments_for_patch(gerrit: Gerrit, patch: Patch):
...
+    logging.info('set_review response = %s', str(gerrit.set_review(change_id=patch.change_id, revision_id=patch.revision_id, review=review)))
```
dlatypov: ditto: can drop str()

git.py
```diff
@@ -18,19 +18,20 @@
...
 import message_dao
 from patch_parser import Patch, Patchset
+from absl import logging
 
 def git(verb: str, *args, cwd=None, input=None) -> str:
-    print(input)
-    print(f'git {verb} ' + ' '.join(args))
-    print(list(args))
+    logging.info('%s', input)
+    logging.info(f'git {verb} ' + ' '.join(args))
```
dlatypov: optional:
```python
logging.info('Running\ngit %s %s\n with input: %s', verb, ' '.join(args), input)
```

git.py
```diff
@@ -18,19 +18,20 @@
...
+    logging.info(list(args))
```
dlatypov: nit: we can drop this log, I think.
It feels redundant with the one above.

We could delete it from the parent commit, but we can have this commit be focused on cleaning up the logging in general.

git.py
```diff
@@ -18,19 +18,20 @@
...
     result = subprocess.run(['git', verb] + list(args),
                             cwd=cwd, input=input,
                             text=True,
                             check = True,
                             stderr=subprocess.STDOUT,
                             stdout=subprocess.PIPE)
     stdout = str(result.stdout)
-    print('stdout: ' + stdout)
+    logging.info('stdout: %s', stdout)
```
dlatypov: might be good to show
```python
logging.info('git %s stdout: %s', verb, stdout)
```

doesn't matter much now since we don't have concurrent calls to this func, it helps to add a bit more context to these things

git.py
```diff
@@ -88,23 +89,21 @@ def shallow_clone(self, depth=1):
...
+        
     def apply_patch(self, patch: Patch):
-        print(patch.text_with_headers)
-        return self._git.am(patch.text_with_headers)
-
+        try:
+            output = self._git.am(patch.text_with_headers)
+            return output
+        except subprocess.CalledProcessError:
+            logging.info('Failed to apply. Aborting...')
```
dlatypov: optional:
```python
logging.warning('Failed to apply patch %s. Aborting...', patch.message_id)
```

archive_updater.py
```diff
@@ -15,44 +15,46 @@
...
 import os
 
+from absl import logging
+
 def fill_message_directory(archive_path: str, directory: str, last_used_commit_hash: str) -> str:
     '''Updates the git repo, then retrieves the MAX_NUMBER_OF_RECENT_COMMITS recent commits and converts them into files stored
-    in the directory corresponding to the passed in directory path.
-
+    in the directory corresponding to the passed in directory path. 
+
```
dlatypov: fyi, looks like some trailing whitespace got added in this patch, including in unrelated lines.
Can you remove it with `sed`?

main.py
```diff
@@ -80,65 +86,73 @@ def update_convert_upload(self):
...
                 gerrit.find_and_label_all_revision_ids(self.gerrit, patchset)
                 gerrit.upload_all_comments(self.gerrit, patchset)
             except Exception as e:
-                print(e)
+                failed += 1
+                logging.error('Failed to upload %s. Gave error: %s', subject, e)
                 continue
-
-    # TODO(@willliu): change print to logging and add logs for # of comments uploaded and # that should be uploaded
+        logging.info('Attempted to upload %d messages but failed to upload %d messsages.', len(messages_to_upload), failed)
```
dlatypov: How about

```python
if failed > 0:
  logging.warning('Failed to upload %d/%d messages', failed, len(messages_to_upload))
```

main.py
```diff
@@ -80,65 +86,73 @@ def update_convert_upload(self):
...
         self.last_hash = archive_updater.fill_message_directory(GIT_PATH, FILE_DIR, self.last_hash)
-        messages = self.archive_index.update(FILE_DIR)
+        messages = self.archive_index.update(FILE_DIR) 
         return messages
-
-    # TODO(@willliu): change print to logging and add logs for # of messages uploaded and # that should be uploaded
+    
     def upload_messages(self, messages_to_upload : List[str]):
+        failed = 0
+        subject = 'Temp Subject'
```
dlatypov:
```suggestion
        subject = '<message not found>`
```

this should make it clearer if the issue was that we couldn't find the patch.

But alternatively, I'd be fine with us just logging the message id instead of the subject.
In this case, they're about as equally useful in looking up the offending email, with the message id being a bit quicker to lookup on lore + the git archives.

main.py
```diff
@@ -80,65 +86,73 @@ def update_convert_upload(self):
...
                 email_thread = self.archive_index.find(message_id)
+                subject = email_thread.subject
                 patchset = patch_parser.parse_comments(email_thread)
                 gerrit.upload_all_comments(self.gerrit, patchset)
             except Exception as e:
-                print(e)
+                failed += 1
+                logging.error('Failed to upload comments for %s. Gave error: %s', subject, e)
                 continue
+        logging.info('Attempted to upload %d comments but failed to upload %d comments.', len(messages_with_new_comments), failed)
```
dlatypov: ditto above

main.py
```diff
@@ -80,65 +86,73 @@ def update_convert_upload(self):
...
+                failed += 1
+                logging.error('Failed to upload %s. Gave error: %s', subject, e)
                 continue
-
-    # TODO(@willliu): change print to logging and add logs for # of comments uploaded and # that should be uploaded
+        logging.info('Attempted to upload %d messages but failed to upload %d messsages.', len(messages_to_upload), failed)
+    
     def upload_comments(self, messages_with_new_comments : List[str]):
+        failed = 0
+        subject = 'Temp Subject'
```
dlatypov: ditto above re subject/message id

patch_parser.py
```diff
@@ -174,7 +175,7 @@ def build_traversal_map(parent_lines: List[Line], child_lines: List[Line], quote
 def find_maximal_map_traversal(traversal_map: Dict[str, List[Line]],
                                parent_lines: List[Line],
                                lines_so_far: List[QuotedLine]) -> List[QuotedLine]:
-    print('len(parent_lines) = ' + str(len(parent_lines)))
+    logging.info('len(parent_lines) = %s', str(len(parent_lines)))
```
dlatypov: `'len(parent_lines) = %d', len(parent_lines)`

patch_parser.py
```diff
@@ -515,12 +516,12 @@ def _parse_patch_file_super_chunk(lines: List[str], raw_index: int) -> List[Patc
         return []
     gerrit_orig_line = int(match.group(1))
     gerrit_new_line = int(match.group(3))
-    print('old starts at: ' + str(gerrit_orig_line) + ', new starts at: ' + str(gerrit_new_line))
+    logging.info('old starts at: %s, new starts at: %s', str(gerrit_orig_line), str(gerrit_new_line))
```
dlatypov: nit: %d and drop the str() casts

patch_parser.py
```diff
@@ -534,7 +535,7 @@ def _parse_patch_file_super_chunk(lines: List[str], raw_index: int) -> List[Patc
 def _parse_patch_file_entry(lines: List[str], index: int) -> Optional[PatchFileLineMap]:
     match = DIFF_LINE_MATCHER.match(lines[0])
     if not match:
-        print('failed to find file diff, instead found: ' + lines[0])
+        logging.info('failed to find file diff, instead found: s', lines[0])
```
dlatypov: %s

patch_parser.py
```diff
@@ -548,29 +549,29 @@ def _parse_patch_file_entry(lines: List[str], index: int) -> Optional[PatchFileL
...
+        logging.info('failed to find ++ b/* line, instead found: %s', lines[0])
         return None
 
     chunks = []
     super_chunk = _parse_patch_file_super_chunk(lines, index)
     while super_chunk:
         chunks.extend(super_chunk)
         chunk = super_chunk[-1]
-        print('parsed super chunk: ' + str(index) + ' to ' + str(chunk.in_range[1]))
+        logging.info('parsed super chunk: %s to %s', str(index), str(chunk.in_range[1]))
```
dlatypov: ditto %d

patch_parser.py
```diff
@@ -634,10 +635,10 @@ def _parse_git_patch(raw_patch: str) -> RawLineToGerritLineMap:
         raise ValueError('Unknown error')
 
 def map_patch_to_gerrit_change(patch: Patch) -> None:
-    print(patch.text)
+    logging.info('Patch: %s', patch.text)
     raw_line_to_gerrit_map = _parse_git_patch(patch.text)
     for comment in patch.comments:
-        print('raw_line: ' + str(comment.raw_line) + ', message: ' + comment.message)
+        logging.info('raw_line: %s, messages: %s', str(comment.raw_line), comment.message)
```
dlatypov: ditto %d, raw_line is an int iirc

patch_parser.py
```diff
@@ -650,7 +651,7 @@ def main():
     map_comments_to_gerrit(patchset)
     for patch in patchset.patches:
         for comment in patch.comments:
-            print('At ' + str(comment.raw_line) + ': ' + str(comment.file) + ':' + str(comment.line) + ':\n' + comment.message)
+            logging.info('At %s: %s: %s:\n %s', str(comment.raw_line), str(comment.file), str(comment.line), comment.message)
```
dlatypov: ditto %d for raw_line

comment.line might be None if there's a logic bug, so perheaps we leave it as %s and drop the cast

dlatypov (changes requested):
Looks a lot nicer!

Can you update the commit description?
I realize you probably just wanted to get it pushed out to get feedback, but wanted to call it out so we don't forget.

Would be good to call out that we're using absl's logging library rather than the standard one.
And if you accept some of my suggestions, that this commit contains some small misc cleanups of the messages themselves.

archive_updater.py
WLIU0628: what is sed?

dlatypov: A CLI utility ("stream editor")
E.g. running this in your shell will get rid of all trailing whitespace
```sh
$ sed -i 's/\s\+$//' *.py
```

main.py
WLIU0628: ok ok, i'll just log the message id then

archive_updater.py
```diff
@@ -39,7 +41,7 @@ def fill_message_directory(archive_path: str, directory: str, last_used_commit_h
     message_hashes = output.decode('utf-8').split()
 
     if len(message_hashes) == 0:
-        print(f'There are no commits in git repo: {archive_path}')
+        logging.warning(f'There are no commits in git repo: {archive_path}')
```
dlatypov: minor: `in git repo: %s', archive_path)`

archive_converter.py
```diff
@@ -67,7 +67,7 @@ def _add_messages_to_index(self, new_messages : List[Message]):
         for message in need_parent:
             parent = self._message_dao.get(message.in_reply_to)
             if not parent:
-                print("Could not find parent email, dropping " + message.subject)
+                logging.info('Could not find parent email, dropping %s', message.subject)
```
dlatypov: optional: perhaps log `message.id` instead to match what we're now doing elsewhere

main.py
```diff
@@ -121,24 +127,30 @@ def upload_messages(self, messages_to_upload : List[str]):
                 gerrit.find_and_label_all_revision_ids(self.gerrit, patchset)
                 gerrit.upload_all_comments(self.gerrit, patchset)
             except Exception as e:
-                print(e)
+                failed += 1
+                logging.error('Failed to upload %s. Gave error: %s', message_id, e)
```
dlatypov: something just to think about:

```python
logging.exception('Failed to upload %s', message_id)
```

This should log the exception along with stack info.
You can use your judgement where an when that would be helpful.

E.g. if you think it could help it make it clearer which of the 3 funcs calls above errored out

patch_parser.py
```diff
@@ -174,7 +175,7 @@ def build_traversal_map(parent_lines: List[Line], child_lines: List[Line], quote
 def find_maximal_map_traversal(traversal_map: Dict[str, List[Line]],
                                parent_lines: List[Line],
                                lines_so_far: List[QuotedLine]) -> List[QuotedLine]:
-    print('len(parent_lines) = ' + str(len(parent_lines)))
+    logging.info('len(parent_lines) = %d', len(parent_lines))
```
dlatypov, line 178: just a note for later: we can consider using `logging.debug()` for a lot of the calls here
By default, things logged at this level wouldn't show up, not even in files.

So that might be something we come back and look at later, in case there are some logs here that turn out to be useful.

Given the file name is included in the log output, it's not too hard to filter out stuff from patch_parser.py

dlatypov (commented):
LGTM!
Apart from two very minor nits and waiting on the parent PR to be merged, this could go in right now.

archive_updater.py
```diff
@@ -15,44 +15,46 @@
...
-    in the directory corresponding to the passed in directory path.
-
+    in the directory corresponding to the passed in directory path. 
+    
     Args:
         archive_path: path to the email archive
         directory: where to store files
         last_used_commit_hash: ending point for filling the directory
-
+
```
bjh83: nit: These new tabs look like a typo.

git.py
```diff
@@ -18,19 +18,18 @@
 import subprocess
 import message_dao
 from patch_parser import Patch, Patchset
+from absl import logging
 
 def git(verb: str, *args, cwd=None, input=None) -> str:
-    print(input)
-    print(f'git {verb} ' + ' '.join(args))
-    print(list(args))
+    logging.info('Running\ngit %s %s\n with input: %s', verb, ' '.join(args), input)
```
bjh83: nit: This looks like it should probably be at the debug level to me.

archive_converter.py
```diff
@@ -67,7 +67,7 @@ def _add_messages_to_index(self, new_messages : List[Message]):
         for message in need_parent:
             parent = self._message_dao.get(message.in_reply_to)
             if not parent:
-                print("Could not find parent email, dropping " + message.subject)
+                logging.info('Could not find parent email, dropping %s', message.id)
```
bjh83: Might want to make a function which prints out the relevant details for dropping an email since you log that message in a couple places.

Probably want to include:
1) message ID as you have done
2) commit hash from lore (since it is hard to search by message ID in lore git repos (if I remember correctly)
3) list the message ID as a web address link which will take you to the email in lore.

dlatypov (approved):

PR #10, opened 2020-11-03
added containerization and made adjustments to git.py to support running in container

WLIU0628:
Added Dockerfile and requirements.txt. The current requirements may have excess libraries and can probably be pruned in the future to limit the image size. Also, as of now, the Dockerfile copies all of src, including gerrit_git_dir if the program was recently run. Please delete gerrit_git_dir and index_files (as described in the Dockerfile) to limit the size of the docker image.


Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# copy the dependencies file to the working directory
+COPY requirements.txt .
+
+# install dependencies
+RUN pip3 install -r requirements.txt
+
+# copy the content of the local src directory to the working directory
+# make sure to delete gerrit_git_dir and index_files from src to limit
+# container image size
```
WLIU0628: should I edit the README.md to add this?

dlatypov: Yes, I'd include this in a section in the README.md as well.

I'd imagine most people would assume initially treat the Dockerfile as a blackbox and not look at it initially.

src/git.py
```diff
@@ -132,4 +142,16 @@ def apply_patchset_and_cleanup(self, patchset: Patchset, messages_dao: message_d
                 self.cleanup_git_dir()
                 raise
 
-
+def main():
```
dlatypov: I assume we can drop this main() now?

I don't think we're expecting people to have issues cloning the repo.

.gitignore
```diff
@@ -3,3 +3,6 @@ token.pickle
 index.pickle
 gerritcookies
 __pycache__
+src/logs
+linux-kselftest
+src/index_files
```
dlatypov, line 12: nit: I'd group the src/ dirs together

.gitignore
```diff
@@ -3,3 +3,6 @@ token.pickle
 index.pickle
 gerritcookies
 __pycache__
+src/logs
```
dlatypov: potentially add a comment above

e.g.
```
# These directories are generated at runtime <maybe elaborate more>
```

dlatypov (commented):
commit description:
Perhaps we can shorten the subject (the initial line) a bit by changing "added containerization" to "Added Dockerfile"

Also, might be worth calling out the issue highlighted by the TODO.
We currently don't have a good identity to use for the committer when running.

Could also make it a bit more readable with some whitespace, e.g.
```
Added Dockerfile and requirements.txt.
The current requirements may have some extra libraries and could be trimmed in the future... .

Note: the Dockerfile copies all of src/, including generated dirs like gerrit_git_dir/.
Please delete gerrit_git_dir to limit the size of the docker image.
```

README.md
```diff
@@ -52,6 +55,21 @@ Then you need to copy the cookies that were added to a file called
...
+## Building a docker image
+
+First, make sure that you have Docker installed on the device which is building
+the image. Before building, there are a couple of folders that should be
+removed if you have run the server locally. These folders include linux-kselftest,
+src/gerrit\_git\_dir, and src/index\_files.
+
+After ensuring these folders are deleted, you can build the image by running the
+following command in the lkml-gerrit-bridge directory:
+``` bash
```
dlatypov: nit:drop space between ` and bash

README.md
```diff
@@ -52,6 +55,21 @@ Then you need to copy the cookies that were added to a file called
 
 NOTE: YOU MUST ADD A COMMENT TO THE TOP OF THE COOKIE FILE: `# HTTP Cookie File`
 
+## Building a docker image
+
+First, make sure that you have Docker installed on the device which is building
+the image. Before building, there are a couple of folders that should be
+removed if you have run the server locally. These folders include linux-kselftest,
```
dlatypov: nit: use `` around the names.

Then you won't have to escape the _

src/git.py
```diff
@@ -92,13 +92,20 @@ def apply_patch(self, patch: Patch):
...
+            logging.warning('Failed to apply patch %s due to %s. Aborting...',
+                            patch.message_id,
+                            e.output)
             git('am', '--abort', cwd=self._git_dir)
             raise
 
     def push_changes(self):
-        return self._git.push('HEAD:refs/for/' + self._branch)
+        try:
+            output = self._git.push('HEAD:refs/for/' + self._branch)
```
dlatypov: we don't need the local variable here
```
try:
   return self._git.push(...)
```

README.md
```diff
@@ -52,6 +55,21 @@ Then you need to copy the cookies that were added to a file called
 
 NOTE: YOU MUST ADD A COMMENT TO THE TOP OF THE COOKIE FILE: `# HTTP Cookie File`
 
+## Building a docker image
```
dlatypov: nit: "###" to match the others

README.md
```diff
@@ -52,6 +55,21 @@ Then you need to copy the cookies that were added to a file called
...
+
+First, make sure that you have Docker installed on the device which is building
+the image. Before building, there are a couple of folders that should be
+removed if you have run the server locally. These folders include linux-kselftest,
+src/gerrit\_git\_dir, and src/index\_files.
```
dlatypov: optional: can intead give a command to copy paste like

```bash
$ rm -rf linux_kselftest/ src/gerrit_git_dir/ src/index_files/
```

dlatypov (commented):
Overall LGTM, just minor nits here and there.

dlatypov (approved):
LGTM.
I'll hold off on merging for now in case Brendan wants to take a look.

bjh83 (approved):
LGTM.

PR #13, opened 2020-11-06
Minor fixes

WLIU0628:
Fixed email_thread typing from str to Message.

Added archive hash to error message from generate_email_from_file.

Changed the REGEX for determining whether an email is a patch. Now
any text enclosed with brackets [] are considered patches.

Fixed a bug with logging and wrote a possible area for future bugs.

Added change_id as a parameter in initializing a Patch.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -122,7 +122,7 @@ class Server(object):
     def upload_messages(self, messages_to_upload : List[str]):
         failed = 0
         for message_id in messages_to_upload:
-            email_thread : str = None
+            email_thread : Message
             try:
                 email_thread = self.archive_index.find(message_id)
                 patchset = patch_parser.parse_comments(email_thread)
@@ -142,7 +142,7 @@ class Server(object):
     def upload_comments(self, messages_with_new_comments : List[str]):
         failed = 0
         for message_id in messages_with_new_comments:
-            email_thread : str = None
+            email_thread : Message
             try:
                 email_thread = self.archive_index.find(message_id)
                 patchset = patch_parser.parse_comments(email_thread)
--- a/src/patch_parser.py
+++ b/src/patch_parser.py
@@ -39,13 +39,13 @@ class CoverLetter(object):
         self.comments = comments
 
 class Patch(object):
-    def __init__(self, message_id, text, text_with_headers, set_index, comments):
+    def __init__(self, message_id, text, text_with_headers, set_index, comments, change_id):
         self.message_id = message_id
         self.text = text
         self.text_with_headers = text_with_headers
         self.set_index = set_index
         self.comments = comments
-        self.change_id = None
+        self.change_id = change_id
         self.revision_id = None
 
 class Patchset(object):
@@ -309,6 +309,7 @@ def find_comments(parent_lines: List[Line], all_child_lines: List[Line]) -> List
     return merge_comment_lines(comment_lines)
 
 def diff_reply(parent: Message, child: Message) -> List[Comment]:
+    #TODO: to_lines only works on str, but Message.content is also sometimes a List
     parent_lines = to_lines(parent.content)
     child_lines = to_lines(child.content)
     return find_comments(parent_lines, child_lines)
@@ -316,7 +317,7 @@ def diff_reply(parent: Message, child: Message) -> List[Comment]:
 def filter_patches_and_cover_letter_replies(email_thread: Message) -> Tuple[List[Message], List[Message]]:
     patches = []
     cover_letter_replies = []
-    if (not email_thread.in_reply_to and parse_set_index(email_thread)[0] == 1):
+    if (not email_thread.in_reply_to and email_thread.patch_index()[0] == 1):
         patches.append(email_thread)
     for message in email_thread.children:
         if message.is_patch():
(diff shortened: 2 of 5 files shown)


src/patch_parser.py
```diff
@@ -309,6 +309,7 @@ def find_comments(parent_lines: List[Line], all_child_lines: List[Line]) -> List
     return merge_comment_lines(comment_lines)
 
 def diff_reply(parent: Message, child: Message) -> List[Comment]:
+    #TODO: Potential bug, to_lines takes a str, but Message.content is a List
```
bjh83: Well, is it a bug? Have you run this code?

bjh83 (commented):
> Fixed email_thread typing from str to Message.
> 
> Added archive hash to error message from generate_email_from_file.
> 
> Changed the REGEX for determining whether an email is a patch. Now
> any text enclosed with brackets [] are considered patches.
> 
> Fixed a bug with logging and wrote a possible area for future bugs.

What do you mean by "a possible area for future bugs?"

> Added change_id as a parameter in initializing a Patch.

src/setup_gmail.py
```diff
@@ -65,7 +65,7 @@ def get_key(self):
             return None
 
     def is_patch(self) -> bool:
-        if re.match(r'\[.+ (\d+)/(\d+)\] .+', self.subject):
+        if re.match(r'\[.+\] .+', self.subject):
```
dlatypov, line 68: can we perhaps move parse_set_index() into Message to keep the two functions closer together?

```python
   def patch_index(self) -> Tuple[int, int]:
      """Returns the patch index, e.g. (1, 6) for the first patch in a series of 6 patches."""
      if not self.is_patch():
         raise ValueError(...)
      ...
```

dlatypov, line 68: Hmm, this can have some false positives.

E.g. `[GIT PULL]`, see https://lore.kernel.org/lkml/20201106173823.48AD72151B@mail.kernel.org/
Maybe we're fine with matching more since it'll just fail to apply and this way we avoid false negatives.

E.g. We could also just match against `'r\[PATCH.*\] .*'`.
This would exclude "[RFC]" and "[RFC PATCH]" (and any other variants thereof).
If we're fine with not including RFCs (I also see someone using "RFC PATCH"

src/main.py
```diff
@@ -122,7 +122,7 @@ def update_message_dir(self) -> List[Message]:
     def upload_messages(self, messages_to_upload : List[str]):
         failed = 0
         for message_id in messages_to_upload:
-            email_thread : str = None
+            email_thread : Message
```
dlatypov, line 125: oh, I'm surprised/sad mypy didn't catch this before?
We did annotate that `parse_comments()` takes a `Message`.

src/setup_gmail.py
bjh83: Good point, maybe we could reuse some of the patch parser code to check if it is a patch? Like this function: https://github.com/google/lkml-gerrit-bridge/blob/cf15a12740d0e54b1c335f9efdbebc4331e2b01f/src/patch_parser.py#L620 ?

Maybe we can create a lightweight version of the patch. Or maybe even just preparse it. That might go beyond the scope of your change here @WLIU0628.

@dlatypov, what do you think?

dlatypov: Hmm.
Having a preparsed `Optional[T]` in `Message` could be a smart way of doing things.

We'd still need to do some regex matching on the subject to grab the patch index in a series so I'm fine with it being left as-is for now in this PR.

We have issue #15 filed now, which we can use to track that further work.

src/patch_parser.py
WLIU0628: the thing is, it works? like comments are uploaded? but sometimes it throws an error?

WLIU0628: like occasionally, there will be an error 'List does not have method splitlines()' or something along those lines

src/main.py
WLIU0628: I think Mypy did catch it but I didn't test it when I made the change because it was in response to a comment I think? IDK it was definitely just my mistake in not checking.

dlatypov: Ack, sg.
I guess that means we can be a bit more confident in the code then if mypy isn't complaining anymore.

And no worries. We really should have setup an automated check w/ mypy so you wouldn't have to remember.
(And ideally, the type annotations would have been correct before you started working on it, but that wasn't the case either :P)

src/setup_gmail.py
```diff
@@ -65,10 +65,18 @@ def get_key(self):
...
             return True
         return False
 
+    def set_index(email: Message) -> Tuple[int, int]:
```
dlatypov:
```suggestion
    def patch_index(self) -> Tuple[int, int]:
```

src/patch_parser.py
dlatypov: Ok, so sounds like we have a random type error where we sometimes assign a str to .content.
I'm fine with leaving this as a TODO here, but perhaps reword to

```suggestion
    #TODO: to_lines only works on str, but Message.content is also sometimes a List.
```

dlatypov (approved):
LGTM.

src/patch_parser.py
bjh83: SGTM.

src/setup_gmail.py
bjh83: SGTM

PR #30, opened 2020-11-18
Add dryrun option and fix some other stuff

bjh83:
Couple of commits here. I can break them up if you want.

First commit adds some more stuff to .gitignore.
Second commit adds a --dryrun option to service, basically runs everything as normal without uploading to Gerrit.
Last commit fixes email decoding.

### DIFF
--- a/src/gerrit.py
+++ b/src/gerrit.py
@@ -40,7 +40,26 @@ class HTTPCookieAuth(AuthBase):
         request.prepare_cookies(self.cookie_jar)
         return request
 
-class Gerrit(object):
+
+class FakeGerrit(object):
+    def new_change(self, change: Dict[str, Any]):
+        pass
+
+    def get_change(self, change_id: str):
+        pass
+
+    def get_patch(self, change_id: str, revision_id: str):
+        pass
+
+    def get_review(self, change_id: str, revision_id: str):
+        pass
+
+    # TODO: fix type checking issues and set revision_id back to str.
+    def set_review(self, change_id: str, revision_id: Any, review: Dict[str, Any]):
+        pass
+
+
+class Gerrit(FakeGerrit):
     def __init__(self, rest_api: GerritRestAPI) -> None:
         self._rest_api = rest_api
 
@@ -68,6 +87,7 @@ class Gerrit(object):
                         change_id=change_id,
                         revision_id=revision_id),
                 data=review)
+
 
 def _find_and_label_revision_id(gerrit: Gerrit, patch: Patch):
     change_info = gerrit.get_change(patch.change_id)
--- a/src/git.py
+++ b/src/git.py
@@ -79,10 +79,10 @@ def _parse_gerrit_patch_push(gerrit_result: str) -> str:
     logging.info('change_id = %s', change_id)
     return change_id
 
-class GerritGit(object):
-    def __init__(self, git_dir: str, cookie_jar_path: str, url: str, project: str, branch: str) -> None:
+
+class DryrunGerritGit(object):
+    def __init__(self, git_dir: str, url: str, project: str, branch: str) -> None:
         self._git_dir = git_dir
-        self._cookie_jar_path = cookie_jar_path
         self._git = _Git(git_dir)
         self._remote = url + '/' + project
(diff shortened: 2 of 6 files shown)


src/message.py
```diff
@@ -14,7 +14,7 @@
 
 import email
 import re
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, BinaryIO
```
dlatypov, line 17: nit: BinaryIO is unused in this commit.

src/git.py
```diff
@@ -128,11 +128,7 @@ def _set_trailers(self, patch: Patch) -> str:
           return self._git.commit('--amend', '-F', f.name)
 
     def push_changes(self) -> str:
-        try:
-            return self._git.push(f'HEAD:refs/for/{self._branch}%notify=NONE')
-        except subprocess.CalledProcessError as e:
-            logging.warning('Failed to push upstream because %s.', e.output)
-            raise
+        pass
```
dlatypov, line 131: Hmm, I would prefer this `DryRunGerritGit` to only contain decls + pass

This probably warrants another commit here first to trim down the interface, e.g. rename `def push_patch()` to `def _push_patch()`

I think the only thing we want to expose is apply_patchsetup_and_cleanup()?

Then the def would become

```python
class DryRunGerritGit:
     def apply_patchsetup_and_cleanup(...):
          pass
```

src/gerrit.py
```diff
@@ -40,7 +40,26 @@ def __call__(self, request: PreparedRequest):
         request.prepare_cookies(self.cookie_jar)
         return request
 
-class Gerrit(object):
+
+class FakeGerrit(object):
```
dlatypov, line 44: nit: Perhaps `DryRunGerrit`

This would more symmetric with GerritGit below (assuming we recapitalize to `DryRunGerritGit`

dlatypov, line 44: nit: I'd omit the `object`, https://google.github.io/styleguide/pyguide.html#39-classes

src/git.py
```diff
@@ -79,10 +79,10 @@ def _parse_gerrit_patch_push(gerrit_result: str) -> str:
     logging.info('change_id = %s', change_id)
     return change_id
 
-class GerritGit(object):
-    def __init__(self, git_dir: str, cookie_jar_path: str, url: str, project: str, branch: str) -> None:
+
+class DryrunGerritGit(object):
```
dlatypov, line 83: nit: perhaps rename to `DryRun`

Also nit: perhaps drop `object`, https://google.github.io/styleguide/pyguide.html#39-classes

dlatypov (changes requested):
Overall LGTM.

The "DryRunClass" as the baseclass seems a bit strange to me, but it works.
Mainly just some nits about the naming there.

src/git.py
dlatypov: I've gone ahead and trimmed the interface down as 5adaf15faf15a33982e915f2cd0b292edd35faec.
So you should just be able to rebase on master.

Might have some merge conflicts (but it felt like those would be unavoidable, even if you made the change yourself)

bjh83: I thought about that, but I kind of like it still trying to apply patches locally and doing some other stuff. I found it to provide some useful feedback.

src/gerrit.py
bjh83: Ah yes, I am still stuck on Python 2

src/git.py
```diff
@@ -167,3 +162,22 @@ def apply_patchset_and_cleanup(self, patchset: Patchset, messages_dao: message_d
...
+        super().__init__(git_dir=git_dir, url=url, project=project, branch=branch)
+
+    def _push_changes(self) -> str:
+        try:
+            return self._git.push(f'HEAD:refs/for/{self._branch}%notify=NONE')
+        except subprocess.CalledProcessError as e:
+            logging.warning('Failed to push upstream because %s.', e.output)
+            raise
+
+    def setup_git_dir(self, clone_depth=1) -> None:
```
dlatypov, line 179: I think this wants to be `_setup_git_dir`

PR #33, opened 2021-01-20
Fix appending trailers to git commits

bjh83:
Previously trailers were not properly added to end of git commit
messages. Apparently, trailers were overwriting the entire message
because the file was not opened and flushed properly.

### DIFF
--- a/src/git.py
+++ b/src/git.py
@@ -121,8 +121,9 @@ class GerritGit(object):
         # Add trailers, changing them if they exist.
         # It's _highly_ unlikely that they'd exist, but this seems to be the
         # most sane way of handling that edge case.
-        with tempfile.NamedTemporaryFile(mode='wt') as f:
+        with tempfile.NamedTemporaryFile(mode='w+') as f:
           f.write(original_message)
+          f.flush() # Flush before we write to the file externally below
           self._git('interpret-trailers', '--in-place', '--if-exists=addIfDifferent',
                     '--trailer', change_id_trailer,  '--trailer', lore_trailer, f.name)
           return self._git.commit('--amend', '-F', f.name)


src/git.py
```diff
@@ -121,8 +121,10 @@ def _set_trailers(self, patch: Patch) -> str:
         # Add trailers, changing them if they exist.
         # It's _highly_ unlikely that they'd exist, but this seems to be the
         # most sane way of handling that edge case.
-        with tempfile.NamedTemporaryFile(mode='wt') as f:
+        with tempfile.NamedTemporaryFile(mode='w+') as f:
+          f.seek(0, os.SEEK_END)
```
dlatypov: I'm not sure what this seek does/if it's necessary.
Perhaps a comment is in order?

src/git.py
```diff
@@ -121,8 +121,10 @@ def _set_trailers(self, patch: Patch) -> str:
...
           f.write(original_message)
+          f.flush()
```
dlatypov: can we have a comment like
```python
  # flush before we write to the file externally below
```

dlatypov:
Commit message has a change-id in it?
Is that intentional?

dlatypov (changes requested):

src/git.py
bjh83: I could have sworn that it didn't work without it, but I just tried it without the seek and it appears to work.

Will fix.

bjh83: Done.

src/git.py
bjh83: Done.

dlatypov (approved):
LGTM.
I also forgot this wasn't submitted.

PR #36, opened 2021-06-30
Created automatic PyTest in GitHub Actions

jaimehisao:
Automatically tests code when pushing and creating a PR in order to prevent untested/not working code from entering the repo's master branch.


requirements.txt
```diff
@@ -2,4 +2,3 @@ absl-py==0.10.0
 pygerrit2==2.0.13
 requests==2.24.0
 requests-oauthlib==1.3.0
-absl-py
```
dlatypov: Hmm, did this get dropped accidentally?
We're still doing a pip install for it.

Also (relevant to a parent commit), if we have a requirements.txt, can we just rely on the `pip install -r` call?

dlatypov (changes requested):
Hmm, I slightly dislike pytest (I instead prefer to use `unittest` directly rather than something that tries to be clever), but this looks like this should all work.

We might want to squash the last two commit since they have the same name. 
Can perhaps leave the "Verified" commits as they are to preserve that status.

requirements.txt
jaimehisao: It wasn't in the original requirements file. The build in GH Actions was breaking because it was missing that module. So I added it in the requirements file (didn't work), so then I removed it and added it as a pip install in the .yml

bjh83 (approved):

jaimehisao:
How should I proceed with the Pytest vs unittest issue? 

Also, originally I was thinking on squashing the whole PR into a single commit, don't know if that affects in any way?

dlatypov:
> How should I proceed with the Pytest vs unittest issue?

Sorry, I was unclear.
Let's go with this. There shouldn't be any difference.

Pytest will just do a bit more magic than we need (looking for plain asserts, etc.)

> 
> Also, originally I was thinking on squashing the whole PR into a single commit, don't know if that affects in any way?

Ok, if you wanted to squash it all when we merge, then yeah, nothing to worry about then.

dlatypov (approved):

PR #37, opened 2021-07-08
Fixed bug where comments where one line off (starter bug)

### LINKED ISSUE #32: Comment get added a few lines below where they should
E.g. https://linux-review.googlesource.com/c/linux/kernel/git/torvalds/linux/+/6958/1/tools/testing/kunit/kunit_parser.py#347

Source email: https://lore.kernel.org/linux-kselftest/CABVgOSnPpC=j7MrcmDpvvG6i_voiFQe4137ieyYX+-9B4=G39w@mail.gmail.com/

The comment should be right on the last changed line, not one line below that.
91ae2f18e30d933b023db4d530231210987d7a5b added a reproducer in the unit tests and fixing the TODOs there should resolve the issue.

Another issue to note: patch_parser.py thinks the trailing quoted section 
```
>         if test_result.status == TestStatus.NO_TESTS:
>                 print(red('[ERROR] ') + yellow('no tests run!'))
>         elif test_result.status == TestStatus.FAILURE_TO_PARSE_TESTS:
>
> base-commit: c4d6fe7311762f2e03b3c27ad38df7c40c80cc93
> --
> 2.29.0.rc1.297.gfa9743e501-goog
>
```

is a pair of empty comments (it picks up the 2 quoted empty lines).

That might be trickier to address and can be looked at separately, in that case.

jaimehisao:
Fixes #32. Worked on the starter bug as requested. Added some more tests on suggestion by @bjh83. 

Turns out that the parser treats modified lines and new lines as the same thing. Making no distinction if the previous line was modified or not. I fixed this by modifying the 'InputSource' object and adding a 'previous_item' attribute to hold the previous item and asses wether the next added line is a modification or not. 

If this solution is correct, there is also an alternative fix which we could do, although I am not sure which one is the more "elegant" fix.


src/patch_parser.py
```diff
@@ -68,18 +74,29 @@ def consume(self, n=1) -> None:
         self._lines = self._lines[n:]
         self._base_line_number += n
 
+    def set_previous_item(self, item):
```
dlatypov: hmm, could we make `consume()` automatically track what the previous line was?
just feels like that could remove a class of potential bugs.

And perhaps we should call it `get_previous_line()` or `previous_line()` to match the terminology used a bit more?
"item" in `__getitem__()` is just for to match python's name for the `[]` operator.

src/patch_parser.py
```diff
@@ -390,28 +428,39 @@ def _parse_patch_file_unchanged_chunk(
...
         gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
     in_start = lines.line_number()
     logging.info('First char - 1: %c', lines[0][0])
     while lines[0] and lines[0][0] == '+':
+        lines.set_previous_item(lines[0])
+        previous = lines.get_previous_item()
         lines.consume()
         gerrit_orig_line += 1
         gerrit_new_line += 1
+        if previous[0] == '-': # maybe add check to see if it is actually a modified one though I think this situation only applies when it actually is modified
```
dlatypov, line 445: I think this might need to be generalize to handle chunks like ++-- and so on.

Perhaps we can pass in the previous chunk into this function.
The loop that calls this builds up a `chunks` list so we should be able to pass in an `Optional[...]` there.

We can refactor these inputs (and return values) into a struct so we're not passing around a bigger and bigger tuple (we maybe should have done that to start with, tbh).

src/patch_parser_test.py
```diff
@@ -70,16 +71,137 @@ def test_parse_with_replies(self):
...
+            --
+                    '''.strip()
+        line_map = patch_parser._parse_git_patch(raw_patch)
+        self.assertEqual(line_map.map(11), ('file', 2), msg=repr(line_map))  # line 2, control assert
+        self.assertEqual(line_map.map(13), ('fileb', 4), msg=repr(line_map))  # line 4 (deleted part due to edition)
+        self.assertEqual(line_map.map(14), ('file', 5), msg=repr(line_map))  # line 4 the modified line
+        self.assertEqual(line_map.map(16), ('file', 7), msg=repr(line_map))  # line 7, control assert
+
+
+    def test_one_deleted_line(self):
```
dlatypov, line 112: I like this, yeah we should have had simpler cases like this.
Really helps isolate the issues, thanks!

dlatypov (commented):
Looks good so far.
I see some tests are still missing in the current version (just have `pass`), so I'll await those being uploaded.

There's also some scattered `print` statements for debug that I assume will get removed soon.
We could convert them to `logging.debug()` calls if you feel they could be useful for people debugging the parser locally. These wouldn't show up by default, so it'd be a good fit.
They're just not quite as easy to see and use as normal prints

jaimehisao:
Removed `test_no_edits` since it isn't testing anything useful. Test `one_edit` is tests the same as `test_one_modified_line`.

src/patch_parser.py
jaimehisao: Done!

.gitignore
```diff
@@ -11,3 +11,7 @@ linux-kselftest
 src/gerrit_git_dir
 src/index_files
 src/logs
+.idea/
```
bjh83: Nit: Could you make these changes in a separate PR?

In this specific case, these config changes are pretty simple, but generally I like to keep changes specific to one logical thing - it makes things easier if you need to do a revert or do a git-blame.

src/patch_parser.py
```diff
@@ -292,7 +326,7 @@ def parse_comments(email_thread: Message) -> Patchset:
             assert length == len(patches)
         text = 'From: {from_}\nSubject: {subject}\n\n{content}'.format(
             from_=patch.from_, subject=patch.subject, content=patch.content)
-        patch_list.append(Patch(message_id = patch.id,
+        patch_list.append(Patch(message_id=patch.id,
```
bjh83: You have **a lot** of formatting fixes here. If it was just a handful, that would be OK, but with as many as there is, it is distracting from the core purpose of this change. Can you submit the formatting fixes as a separate change?

src/patch_parser.py
```diff
@@ -366,20 +401,24 @@ def __repr__(self) -> str:
...
     line = lines[0]
     return (line == '--') or (len(lines) <= 1) or bool(SKIP_LINE_MATCHER.match(line) or DIFF_LINE_MATCHER.match(line))
 
+
 def _parse_patch_file_unchanged_chunk(
         lines: InputSource,
         gerrit_orig_line: int,
         gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
     in_start = lines.line_number()
-    while (not _does_match_end_of_super_chunk(lines)) and ((not lines[0]) or (lines[0] and lines[0][0] != '+' and lines[0][0] != '-')):
```
bjh83: Did you change the logic here at all?

src/patch_parser.py
```diff
@@ -390,22 +429,34 @@ def _parse_patch_file_unchanged_chunk(
...
 def _parse_patch_file_added_chunk(
         lines: InputSource,
         gerrit_orig_line: int,
         gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
     in_start = lines.line_number()
     logging.info('First char - 1: %c', lines[0][0])
     while lines[0] and lines[0][0] == '+':
+        previous = lines.get_previous_line()
+        print('ooc previous', previous)
+        print('ooc Item 1 -> ', lines[0])
```
bjh83: These look like debugging lines. If they are, remove them. If not, please log them.

src/patch_parser.py
bjh83: +1

src/patch_parser_test.py
bjh83: +1 Nice job!

jaimehisao:
This PR as it covers the one line off bug only, it makes sense to split this fix from another more complex one (the multiple edited lines one). Will open new PR that will contain the new changes that address the multi-line bug.

PR #41, opened 2021-07-21
Fixing trailing line issue

jaimehisao:
There's an issue that caused lines to be added to the last chunk when multiple lines were deleted, this has been addressed with this PR, and tests have been added to make sure of it.


jaimehisao:
Worth mentioning git issues regarding multiple unrelated commits being added has been fixed, we are now tracking master from google repo instead of my fork.

src/patch_parser.py
```diff
@@ -82,6 +85,15 @@ def set_previous_line(self, item):
     def get_previous_line(self):
         return self._previous_item
 
+    def get_number_of_deleted_lines(self):
```
dlatypov: Suggestion:

Let's consider adding something like

```python
import dataclasses

@dataclasses.dataclass
class HunkParserState:
  gerrit_orig_line: int
  gerrit_new_line: int
  deleted_lines: int
```

Then we can pass that around the parse hunk functions instead of taking multiple arguments and returning updated versions of all of them.

And that gives us a place to store deleted_lines().
It feels like we're using `InputSource` as a handy place to store this additional state (`deleted_lines`) bc the Tuple was getting unwieldy.
But it doesn't really feel like it belongs, given the more generic nature of InputSource.

src/patch_parser.py
```diff
@@ -423,21 +441,33 @@ def _parse_patch_file_unchanged_chunk(
         lines.consume()
         gerrit_orig_line += 1
         gerrit_new_line += 1
-    return (gerrit_orig_line,
+        #logging.warning('Unchanged start: ' + str(in_start))
```
dlatypov: we might want these to keep debugging while refactoring, but reminder to go back and delete these before we merge

src/patch_parser.py
```diff
@@ -423,21 +441,33 @@ def _parse_patch_file_unchanged_chunk(
...
+        #logging.warning('Unchanged line: ' + lines[0])
+        #logging.warning('Unchanged lines - 1: ' + str(lines.line_number() - 1))
+    if lines.get_number_of_deleted_lines() == 0:
```
dlatypov: Aren't these branches effectively the same?

i.e. `gerrit_new_line - lines.line_number() == gerrit_new_line - lines.get_number_of_deleted_lines() - lines.line_number()` when `get_number_of_deleted_lines() == 0`.

Perhaps just adding a new local variable would work better

```
offset = gerrit_new_line - lines.get_number_of_deleted_lines() - lines.line_number()
return ....
```

src/patch_parser.py
```diff
@@ -473,14 +504,14 @@ def _parse_patch_file_removed_chunk(
 
 def _parse_patch_file_chunk(lines: InputSource,
                             gerrit_orig_line: int,
-                            gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
+                            gerrit_new_line: int, chunks: PatchFileChunkLineMap) -> Tuple[int, int, PatchFileChunkLineMap]:
```
dlatypov: we don't seem to be using the newly added chunks parameter here or in `__parse_patch_file_added_chunk()`.
I assume it's a leftover from before when you were considering inspecting the previous hunk to find `deleted_lines()`?

src/patch_parser.py
```diff
@@ -352,6 +364,12 @@ def map(self, raw_line: int) -> Tuple[str, int]:
             raise IndexError(
                 'Expected ' + str(self.in_range[0]) + ' <= ' + str(raw_line) + ' <= ' + str(self.in_range[1]))
 
+    def get_range(self) -> Tuple[int, int]:
+        return self.in_range
+
+    def get_side(self) -> str:
+        return self.side
```
dlatypov: these new functions seem to be unused now.
I assume this is from a previous attempt at inspecting the chunks map to find `deleted_lines` instead of storing it in InputSource.

So I think we can drop these for now.

src/patch_parser_test.py
```diff
@@ -136,6 +136,43 @@ def test_one_deleted_line(self):
...
+                     line 10  (then 6) # this is line 19 in raw_patch
+
+                    base-commit: 235360eb7cd778d7264c5e57358a3d144936b862
+                    --
+                            '''.strip()
+        line_map = patch_parser._parse_git_patch(raw_patch)
+        self.assertEqual(line_map.map(11), ('file', 2), msg=repr(line_map))  # line 2 (11), control assert
+        self.assertEqual(line_map.map(12), ('file', 3), msg=repr(line_map))  # line 3 (12), control assert
+        self.assertEqual(line_map.map(13), ('fileb', 4), msg=repr(line_map))  # Lines 13 - 16 are deleted and should
+        self.assertEqual(line_map.map(14), ('fileb', 5), msg=repr(line_map))  # show on 'fileb' with their corresponding
```
dlatypov: optional: you seem to have line-wrapped the comments, but this one word sticks out.
I think we can just move it into the comment below and everything should fit.

dlatypov (changes requested):

src/patch_parser.py
```diff
@@ -82,6 +85,15 @@ def set_previous_line(self, item):
     def get_previous_line(self):
         return self._previous_item
 
+    def get_number_of_deleted_lines(self):
+        return self._deleted_lines
+
+    def add_deleted_line(self):
+        self._deleted_lines = self._deleted_lines + 1
```
bjh83: I agree with what Daniel said above, so you might move this out or get rid of it entirely. Nevertheless, you could make this:

```
self._deleted_lines += 1
```

src/patch_parser.py
jaimehisao: That's right, they are un-used. Let me drop them as well as the other un-used items.

bjh83 (commented):
Other than Daniel's comments, LGTM.

src/patch_parser.py
jaimehisao: Working on this...

src/patch_parser.py
jaimehisao: This is very true! Good catch, I went ahead and changed it.

src/patch_parser.py
```diff
@@ -415,34 +433,41 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
...
     logging.info('First char - 1: %c', lines[0][0])
     while lines[0] and lines[0][0] == '+':
-        previous = lines.get_previous_line()
         previous = lines.get_previous_line()
         lines.consume()
         gerrit_orig_line += 1
         gerrit_new_line += 1
-        if previous[0] == '-': # maybe add check to see if it is actually a modified one though I think this situation only applies when it actually is modified
+        if previous[
+            0] == '-':  # maybe add check to see if it is actually a modified one though I think this situation only applies when it actually is modified
```
bjh83: Nit: Can you make this a TODO comment and make it a wrapped line comment? This line is pretty long.

src/patch_parser.py
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import dataclasses
```
dlatypov, line 15: very minor nit: can we add a \n before this so there's a separation between license comment and the code?

src/patch_parser.py
```diff
@@ -512,7 +541,7 @@ def _parse_patch_file_super_chunk(lines: InputSource) -> List[PatchFileChunkLine
          gerrit_new_line,
          chunk) = _parse_patch_file_chunk(lines,
                                           gerrit_orig_line,
-                                          gerrit_new_line)
+                                          gerrit_new_line, parser_state)
```
bjh83: Can you put this on a new line to match the rest of the parameters here?

src/patch_parser.py
```diff
@@ -156,6 +156,24 @@ def diff_best_match(self, string: List[str]) -> List[str]:
         return node.diff_best_match(string)
 
 
+@dataclasses.dataclass
+class HunkParserState:
+    _number_of_edits: int
+    _deleted_lines: int
+
+    def get_number_of_deleted_lines(self):
```
dlatypov: nit: I'd drop the getters and setters here since we're using a dataclass.
We can have the callers interact with the fields directly, or use a helper where it makes sense (e.g. if we move gerrit_{orig,new}_line in here, then we could have a helper to increment them all)

src/patch_parser.py
```diff
@@ -156,6 +156,24 @@ def diff_best_match(self, string: List[str]) -> List[str]:
         return node.diff_best_match(string)
 
 
+@dataclasses.dataclass
+class HunkParserState:
+    _number_of_edits: int
```
dlatypov: nit: this field seems unused, let's drop it

src/patch_parser.py
```diff
@@ -415,34 +433,41 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
...
     in_start = lines.line_number()
     while (not _does_match_end_of_super_chunk(lines)) and (
             (not lines[0]) or (lines[0] and lines[0][0] != '+' and lines[0][0] != '-')):
         logging.info('dropping line: %s', lines[0])
         lines.consume()
         gerrit_orig_line += 1
         gerrit_new_line += 1
+        # logging.warning('Unchanged start: ' + str(in_start))
+        # logging.warning('Unchanged line: ' + lines[0])
+        # logging.warning('Unchanged lines - 1: ' + str(lines.line_number() - 1))
```
bjh83: As I think Daniel mentioned, please drop these comments when you are done or promote them to debug log statements.

src/patch_parser.py
```diff
@@ -156,6 +156,24 @@ def diff_best_match(self, string: List[str]) -> List[str]:
...
+    def get_number_of_deleted_lines(self):
+        return self._deleted_lines
+
+    def add_deleted_line(self):
+        self._deleted_lines += 1
+
+    def reset_deleted_lines(self):
+        self._deleted_lines = 0
+
+    def __init__(self):
```
dlatypov: we don't need this, I think.

But we should be able to also do
```python
deleted_lines: int = 0 
```
up top to be more explicit instead of needing an `__init__()`

src/patch_parser.py
```diff
@@ -415,34 +433,41 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
 def _parse_patch_file_unchanged_chunk(
         lines: InputSource,
         gerrit_orig_line: int,
-        gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
+        gerrit_new_line: int,
```
dlatypov: I was originally thinking we could move these gerrit_orig/new_line fields into HunkParserState.
Wdyt?

src/patch_parser.py
dlatypov: the added line wrapping here seems weird.
Perhaps let's revert this hunk?

src/patch_parser.py
```diff
@@ -458,9 +483,11 @@ def _parse_patch_file_added_chunk(
 def _parse_patch_file_removed_chunk(
         lines: InputSource,
         gerrit_orig_line: int,
-        gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
+        gerrit_new_line: int,
+        parser_state: HunkParserState) -> Tuple[int, int, PatchFileChunkLineMap]:
     in_start = lines.line_number()
     while lines[0] and lines[0][0] == '-':
+        parser_state.add_deleted_line()
```
dlatypov: nit: I'd slightly prefer the `lines.consume()` comes first and this integer increment goes down with the rest (`gerrit_orig_line += 1`)

dlatypov (changes requested):

src/patch_parser.py
jaimehisao: huh I hadn't noticed that!

src/patch_parser.py
jaimehisao: I see, to do everything in one go basically?

src/patch_parser.py
jaimehisao: It's fine, I just left them there because of the original design, but I can move them over.

src/patch_parser.py
dlatypov: re moving new fields in: yeah, I was thinking it'd be fine to do that in one go.
But if you'd prefer, we could do this separately.

One option could be to do this in two PRs, and another is to clean up the commits a bit so we merge this PR w/o squashing.
I.e. squash all the current commits, and then have a second commit that moves the remaining fields into `HunkParserState`

src/patch_parser.py
```diff
@@ -414,105 +422,97 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
...
+                                 side='', offset=offset)
 
 
 def _parse_patch_file_added_chunk(
         lines: InputSource,
-        gerrit_orig_line: int,
-        gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
+        parser_state: HunkParserState) -> PatchFileChunkLineMap:
     in_start = lines.line_number()
+    previous_number_of_removed_lines = parser_state.deleted_lines
```
dlatypov: this var is currently unused, did you want to use it to compute offset?
i.e.
offset=(parser_state.gerrit_new_line - lines.line_number() - previous_deleted_lines
?

or that can be left for another commit and we can drop this var for now.

src/patch_parser.py
```diff
@@ -414,105 +422,96 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
...
-                            gerrit_new_line: int) -> Tuple[int, int, PatchFileChunkLineMap]:
+                            parser_state: HunkParserState) -> PatchFileChunkLineMap:
     line = lines[0]
     start_line_len = len(lines)
     if _does_match_end_of_super_chunk(lines):
         raise ValueError('Unexpected line: ' + line)
     elif line and line[0] == '+':
         logging.info('First char - 0: %c', line[0])
-        ret_val = _parse_patch_file_added_chunk(lines, gerrit_orig_line, gerrit_new_line)
+        ret_val = _parse_patch_file_added_chunk(lines, parser_state)
```
dlatypov: nit: while we're here, can we update ret_val to something better?

It was called that since it was a tuple and didn't have a good name.
perhaps `chunk_map` ?

src/patch_parser.py
```diff
@@ -414,105 +422,96 @@ def _does_match_end_of_super_chunk(lines: InputSource) -> bool:
...
-        gerrit_orig_line += 1
-        gerrit_new_line += 1
-    return (gerrit_orig_line,
-            gerrit_new_line,
-            PatchFileChunkLineMap(in_range=(in_start, lines.line_number() - 1),
-                                  side='',
-                                  offset=(gerrit_new_line - lines.line_number())))
+        parser_state.gerrit_orig_line += 1
+        parser_state.gerrit_new_line += 1
+        # logging.debug('Unchanged start: ' + str(in_start))
```
dlatypov: reminder to drop these commented out logs.

dlatypov (changes requested):
LGTM, just would want to clean it up a bit more.

jaimehisao:
Code has been cleaned up!

dlatypov (approved):
LGTM, thanks!

PR #43, opened 2021-07-26
Codecov code upload fails when pushing from a fork

jaimehisao:
Codecov fails when running in a fork, causing an email to be sent out with the failed run. This is supposed to remedy that. Mostly focused on reducing email spam to developers.


dlatypov (approved):
Looks like the documentation is https://github.com/codecov/codecov-action#arguments

It says "Specify if CI pipeline should fail when Codecov runs into errors during upload."
So I guess this change is focused on reducing email spam, which sounds good to me.

bjh83 (approved):

PR #46, opened 2022-06-17
Fix is_patch() so it returns False on cover letters

bjh83:
is_patch() is a method on Message objects that we use to determine
whether we should treat an email message as a patch or not. Currently,
it returns True for the cover letter of a patch set - fix this.

Signed-off-by: Brendan Higgins <brendanhiggins@google.com>

### DIFF
--- a/src/message.py
+++ b/src/message.py
@@ -31,14 +31,25 @@ class Message(object):
         self.archive_hash = archive_hash
         self.children = []  # type: List[Message]
 
-    def is_patch(self) -> bool:
+    def _is_patch_or_coverletter(self) -> bool:
         if re.match(r'\[.+\] .+', self.subject):
             return True
         return False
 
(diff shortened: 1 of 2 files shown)


src/message.py
```diff
@@ -31,14 +31,25 @@ def __init__(self, id, subject, from_, in_reply_to, content, archive_hash) -> No
         self.archive_hash = archive_hash
         self.children = []  # type: List[Message]
 
-    def is_patch(self) -> bool:
+    def is_patch_or_coverletter(self) -> bool:
```
dlatypov: very nit: perhaps `_is_patch_or_coverletter()` since I don't think we'll see much reason for its use outside this file?

dlatypov (approved):
Some optional suggestions below, but otherwise LGTM.

PR #48, opened 2022-07-07
Provide a more step-by-step instruction for getting started

lenhard-thomas:


README.md
```diff
@@ -28,8 +28,10 @@ also need to build a docker image.
 
 First you need to get Gerrit to generate an API cookie; you can do this by:
 
-- Going to your Gerrit's webpage
-- Clicking on "Generate Password"
+- Going to your Gerrit's dashboard
```
dlatypov: this is very much a nit, but I'd prefer the imperative tense now we have more steps.
i.e.
```
First you need to get Gerrit to generate an API cookie. To get one:
- Go to your Gerrit instance's dashboard
- Click on "Settings"
...
```

dlatypov: nit: perhaps include an example like "(e.g. https://linux-review.googlesource.com/)" or maybe directly link to https://linux-review.googlesource.com/settings ?

dlatypov (commented):
As noted by the failing check, we'll need to figure out the CLA situation before you can submit any code.

Also, some minor nits/suggestions

README.md
lenhard-thomas: Definitely agree! Updated the steps to be in imperative tense.

lenhard-thomas: Added a direct link.

dlatypov (approved):
Looks good, thanks!

I forgot the code relied on a "./gerritcookies" here, https://github.com/google/lkml-gerrit-bridge/blob/bb4052b300736fbcf87d220b1ff83b559b9702e1/src/main.py#L36
So the dockerfile change also makes sense to me.

dlatypov:
Actually, I forgot the look at the commit description.
Since this is a single commit, we'll probably just "rebase and merge", so it would be nice if the commit description could explain the dockerfile change.

It can be something short like 1bde2f5e05be601ca33ec3697df6a298f426d204.
Note: you don't need the "Signed-off-by" line like that has. That's just force of habit on Brendan's part from LKML.

bjh83 (approved):

PR #49, opened 2022-07-13
Database

lenhard-thomas:
Previously, messages would be stored into the database immediately after being processed. Now, instead, a dictionary (local_mapping) is used to handle the intermediate processing that occurs before pushing patches onto Gerrit, thus the database stores patches only once a patch has been posted onto Gerrit. 

There was an issue within git.py where the GERRIT_PUSH_MATCHER constraint was too strong such that some pushes weren't being matched properly. Thus, it was modified to look for just the `success` portion of the push result as this is the only part that's needed to determine the change url.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 # set base image (host OS)
 FROM python:3.8
 
+# allow Google API to utilize downloaded credentials
+ENV GOOGLE_APPLICATION_CREDENTIALS credentials.json
+
 # set the working directory in the container
 WORKDIR /code
 
@@ -9,6 +12,9 @@ COPY requirements.txt .
 
 # copy the gerrit cookies file to the working directory
 COPY gerritcookies .
+
+# copy the gcloud credentials file to the working directory
+COPY credentials.json .
 
 # install dependencies
 RUN pip3 install -r requirements.txt
--- a/README.md
+++ b/README.md
@@ -49,6 +49,19 @@ you have to:
 Then you need to copy the credentials file that was added to a file called 
 `credentials.json` in this directory.
 
+### Setting up Database information
+
+Create a file named `.env` under `src/` that includes all of the following:
+
+```
+HOST = [HOST_NAME]
+USER = [USER_NAME]
+PASSWORD = [PASSWORD]
+DB = [NAME_OF_DATABASE]
+
+NOTE: The name of the database does not have to be an existing database.
+```
+
 ### Building a docker image
 
 First, make sure that you have Docker installed on the device which is building
--- a/README.md
+++ b/README.md
@@ -38,6 +38,17 @@ Then you need to copy the cookies that were added to a file called
 
 NOTE: YOU MUST ADD A COMMENT TO THE TOP OF THE COOKIE FILE: `# HTTP Cookie File`
 
+### Setting up Google Cloud credentials
+
+First you need to generate the gcloud credentials file. In order to do that,
+you have to:
+
+- Download gcloud CLI (follow instructions on https://cloud.google.com/sdk/docs/install)
+- Run `gcloud auth appplication-default login` in the terminal
+
+Then you need to copy the credentials file that was added to a file called 
+`credentials.json` in this directory.
+
 ### Building a docker image
 
 First, make sure that you have Docker installed on the device which is building
--- a/src/archive_converter.py
+++ b/src/archive_converter.py
@@ -21,55 +21,47 @@ from message import Message, parse_message_from_str
 from message_dao import MessageDao
 
 class ArchiveMessageIndex(object):
-    def __init__(self, message_dao : MessageDao) -> None:
+    def __init__(self, message_dao: MessageDao) -> None:
         self._message_dao = message_dao
 
-    def update(self, data_dir: str) -> List[Message]:
+    def update(self, data_dir: str) -> Dict[str, Message]:
         """ Updates index with messages in the passed in directory.
-        Returns a list of new messages """
+        Returns a dictionary mapping new messages' ids to their corresponding message."""
 
-        new_messages : List[Message] = []
+        new_messages : Dict[str, Message] = {}
 
         for filename in os.listdir(data_dir):
             if not filename.endswith(".txt"):
                 continue
             email = generate_email_from_file(os.path.join(data_dir, filename))
             if email and not self._message_dao.get(email.id):
-                new_messages.append(email)
-        self._add_messages_to_index(new_messages)
+                new_messages[email.id] = email
+        self._populate_children(new_messages)
         return new_messages
 
-    def size(self) -> int:
-        return self._message_dao.size()
+    def _populate_children(self, new_messages: Dict[str, Message]) -> None:
+        """ Iterates through all new emails and links together emails that form
+        a thread by populating message.children. Parents from the database
+        that have a new reply are added to 'new_messages' in order to be stored
+        in the database once re-uploaded."""
 
-    def find(self, message_id : str) -> Message:
-        message = self._message_dao.get(message_id)
-        if message is None:
-            raise ValueError(f'Could not find message: {message_id}')
-        return message
+        # Iterate through a copy of the values to avoid altering the size of the iterator
+        for message in list(new_messages.values()):
 
-    def _add_messages_to_index(self, new_messages : List[Message]) -> None:
-        """ Iterates through all new emails and links together emails that form a thread by populating message.children. """
-
-        need_parent : List[Message] = []
-
-        # First iterates through all messages to distinguish between 1.) replies and 2.) the start of a thread.
-        for message in new_messages:
-            if not self._message_dao.get(message.id):
-                self._message_dao.store(message)
-
-            # If message is a reply, it needs to be associated with an existing thread
-            if message.in_reply_to is not None:
-                need_parent.append(message)
+            # Associate a message with an existing thread only if the message is a reply
+            if message.in_reply_to is None:
                 continue
 
-        for message in need_parent:
-            parent = self._message_dao.get(message.in_reply_to)
-            if not parent:
-                logging.info('Could not find parent email, dropping %s', message.debug_info())
-                continue
-            parent.children.append(message)
-            self._message_dao.store(parent)
+            if message.in_reply_to in new_messages:
+                parent = new_messages.get(message.in_reply_to)
+                parent.children.append(message)
+            else:
+                parent = self._message_dao.get(message.in_reply_to)
+                if not parent:
+                    logging.info('Could not find parent email, dropping %s', message.debug_info())
+                    continue
+                parent.children.append(message)
+                new_messages[parent.id] = parent
 
 def generate_email_from_file(file: str) -> Optional[Message]:
     archive_hash = file[12:-4]
--- a/src/git.py
+++ b/src/git.py
@@ -19,8 +19,8 @@ import shutil
 import subprocess
 import tempfile
 
-from message import lore_link
 import message_dao
+from message import lore_link, Message
 from patch_parser import Patch, Patchset
 from absl import logging
 
@@ -58,15 +58,15 @@ class _Git(object):
     def commit(self, *args) -> str:
         return _git('commit', *args, cwd=self._git_dir)
 
-GERRIT_PUSH_MATCHER = re.compile(
-r'remote: Processing changes: refs: 1, new: 1, done(?:\s+remote: (?:(?:commit)|(?:warning)).+$)?\s+remote:\s+remote: SUCCESS\s+remote:\s+remote:\s+(https://[\w/+.-]+)\s+',
+GERRIT_CHANGE_URL_MATCHER = re.compile(
+r'SUCCESS\s+remote:\s+remote:\s+(https://[\w/+.-]+)\s+',
 flags=re.MULTILINE)
 
 GERRIT_CHANGE_ID_MATCHER = re.compile(r'^https://[\w/+.-]+\+/(\d+)$')
 
 def _parse_gerrit_patch_push(gerrit_result: str) -> str:
     logging.info('%s', gerrit_result)
-    match = GERRIT_PUSH_MATCHER.search(gerrit_result)
+    match = GERRIT_CHANGE_URL_MATCHER.search(gerrit_result)
     if match is None:
       raise ValueError(f'Could not find change url from gerrit output: {gerrit_result}')
     change_url = match.group(1)
@@ -154,17 +154,16 @@ class GerritGit(object):
     def _cleanup_git_dir(self) -> None:
         shutil.rmtree(self._git_dir)
 
-    # Pass in the dao so that patches can be updated when they are pushed, this way less lost data when an error happens
-    def apply_patchset_and_cleanup(self, patchset: Patchset, messages_dao: message_dao.MessageDao):
+    # Pass in the dao so that patches can be updated when they are pushed, this way less lost data when an error happens,
+    # and pass in the message directly to minimize database lookups
+    def apply_patchset_and_cleanup(self, patchset: Patchset, message: Message, message_dao: message_dao.MessageDao):
         if not os.path.isdir(self._git_dir):
             self._setup_git_dir()
         for patch in patchset.patches:
             # This should cause the server to clean up the git dir because of potential failures
             try:
-                message = messages_dao.get(patch.message_id)
-                if message:
-                    message.change_id = self._push_patch(patch).change_id
-                    messages_dao.store(message)
+                message.change_id = self._push_patch(patch).change_id
+                message_dao.store(message)
             except:
                 self._cleanup_git_dir()
                 raise
--- a/src/message.py
+++ b/src/message.py
@@ -24,6 +24,7 @@ class Message(object):
     def __init__(self, id, subject, from_, in_reply_to, content, archive_hash) -> None:
         self.id = id
         self.subject = subject
+        self.normalized_subject = self._normalize_subject()
         self.from_ = from_
         self.in_reply_to = in_reply_to
         self.content = content
@@ -55,6 +56,9 @@ class Message(object):
             return int(match.group(1)), int(match.group(2))
         return (1, 1)
 
+    def _normalize_subject(self) -> str:
+        return self.subject.partition("] ")[2].lower()
+
     def __str__(self) -> str:
         in_reply_to = self.in_reply_to or ''
         return ('{\n' +
@@ -66,6 +70,14 @@ class Message(object):
 
     def __repr__(self) -> str:
         return str(self)
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, Message):
+            return False
+        return (self.id == other.id and self.subject == other.subject
+        and self.from_ == other.from_ and self.in_reply_to == other.in_reply_to
+        and self.content == other.content and self.change_id == other.change_id
+        and self.archive_hash == other.archive_hash and self.children == other.children)
 
     def debug_info(self) -> str:
         return (f'Message ID: {self.id}\n'
--- a/src/message_dao.py
+++ b/src/message_dao.py
@@ -12,27 +12,139 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, Optional
-from message import Message
+import json
+import os
+import subprocess
 
+from functools import lru_cache
+from typing import Dict, List, Optional
+
+from dotenv import load_dotenv
+from google.cloud.sql.connector import Connector
+from message import lore_link, Message, parse_message_from_str
+
+load_dotenv()
 EPOCH_HASH = 'ae9e7be4a03765456fe38287533e6446e8bbc93c'
 
 class MessageDao(object):
+    def __init__(self, archive_path: str) -> None:
+        """ Creates a connection as well as two tables: Messages and States.
+        Message stores the messages we've uploaded and States is a key-value
+        store which tracks things like 'last_hash', the last Lore git commit
+        we've processed."""
+        self._initialize_connection()
+        self._initialize_tables()
+        self.archive_path = archive_path
 
+    def _initialize_connection(self) -> None:
+        connector = Connector()
+        self.connection = connector.connect(
+            os.environ.get("HOST"),
+            "pymysql",
+            user = os.environ.get("USER"),
+            password = os.environ.get("PASSWORD")
+        )
+    
+    def _initialize_tables(self) -> None:
+        db_name = os.environ.get("DB")
+        if not db_name:
+            raise Exception("Missing environment variable for name of database.")
+        with self.connection.cursor() as cursor:
+            cursor.execute("CREATE DATABASE IF NOT EXISTS " + db_name)
+            self.connection.select_db(db_name)
+            # Mapping from message id to message
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS Messages"
+                "(message_id VARCHAR(255) NOT NULL,"
+                "normalized_subject VARCHAR(255) NOT NULL,"
+                "from_ VARCHAR(255) NOT NULL,"
+                "in_reply_to VARCHAR(255),"
+                "archive_hash VARCHAR(255) NOT NULL,"
+                "change_id VARCHAR(255),"
+                "lore_link VARCHAR(255),"
+                "PRIMARY KEY (message_id))"
+            )
+            # Mapping from name of state attribute to state
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS States"
+                "(state_name VARCHAR(255) NOT NULL,"
+                "value VARCHAR(255) NOT NULL,"
+                "PRIMARY KEY (state_name))"
+            )
+        self.connection.commit()
+
+    def store(self, message: Message) -> None:
+        link = lore_link(message.id)
+        query = "REPLACE INTO Messages VALUES (%s, %s, %s, %s, %s, %s, %s)"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, (message.id, message.normalized_subject, message.from_,
+            message.in_reply_to, message.archive_hash, message.change_id, link))
+        if message.in_reply_to:
+            # Clear cache because the parent's cache is no longer valid: list of children changed
+            self.get.cache_clear()
+        self.connection.commit()
+
+    def _get_children(self, message_id: str) -> List[Optional[Message]]:
+        query = "SELECT * FROM Messages WHERE in_reply_to=%s"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, (message_id,))
+            res = cursor.fetchall()
+        return [self.get(tup[0]) for tup in res]
+
+    @lru_cache
+    def get(self, message_id: str) -> Optional[Message]:
+        query = "SELECT archive_hash, change_id FROM Messages WHERE message_id=%s"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, (message_id,))
+            res = cursor.fetchone()
+        if res is None:
+            return None
+        archive_hash, change_id = res[0], res[1]
+        # Recreate the message object using the archive hash
+        raw_email = subprocess.check_output(['git', '-C', self.archive_path, 'show', f'{archive_hash}:m'])
+        msg = parse_message_from_str(raw_email.decode(), archive_hash=archive_hash)
+        msg.change_id = change_id
+        msg.children = self._get_children(message_id)
+        return msg
+
+    def size(self) -> int:
+        query = "SELECT COUNT(*) FROM Messages"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query)
+            res = cursor.fetchone()
+        return res[0]
+
+    def store_last_hash(self, last_hash: str) -> None:
+        query = "REPLACE INTO States VALUES (%s, %s)"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, ("last_hash", last_hash))
+        self.connection.commit()
+
+    def get_last_hash(self) -> str:
+        query = "SELECT value FROM States WHERE state_name=%s"
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, ("last_hash"))
+            res = cursor.fetchone()
+        return EPOCH_HASH if res is None else res[0]
+
+
+class FakeMessageDao(MessageDao):
     def __init__(self) -> None:
         # Maps message.id to message
-        self._messages_seen : Dict[str, Message] = {}
+        self._messages_seen = {}
+        self.last_hash = EPOCH_HASH
 
-    def store(self, message : Message) -> None:
+    def store(self, message: Message) -> None:
         self._messages_seen[message.id] = message
 
-    def get(self, message_id : str) -> Optional[Message]:
+    def get(self, message_id: str) -> Optional[Message]:
         return self._messages_seen.get(message_id)
 
     def size(self) -> int:
         return len(self._messages_seen)
 
-    # This function is currently hardcoded because it will not be used until post database integration.
-    # TODO (@willliu): if empty, then return epoch hash. o/w return latest set hash.
-    def get_last_hash(self) -> str:
-        return EPOCH_HASH
+    def store_last_hash(self, last_hash: str) -> None:
+        self.last_hash = last_hash
+
+    def get_last_hash(self) -> int:
+        return self.last_hash
--- a/src/message_dao_test.py
+++ b/src/message_dao_test.py
@@ -1,0 +1,88 @@
+import unittest
+import subprocess
+import zlib
+
+from unittest import mock
+from google.cloud.sql.connector import Connector
+
+import message
+import message_dao
+import archive_converter
+from test_helpers import test_data_path
+
+class StrContains(str):
+   def __eq__(self, other):
+      return self in other
+
+class MessageDaoTest(unittest.TestCase):
+
+    def setUp(self):
+        self.addCleanup(mock.patch.stopall)
+        self.mock_connect = mock.patch.object(Connector, 'connect').start()
+        mock_connection = self.mock_connect.return_value
+        self.mock_commit = mock_connection.commit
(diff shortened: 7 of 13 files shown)


README.md
```diff
@@ -37,6 +37,16 @@ Then you need to copy the cookies that were added to a file called
...
 NOTE: YOU MUST ADD A COMMENT TO THE TOP OF THE COOKIE FILE: `# HTTP Cookie File`
 
+### Setting up Google Cloud credentials
+
+First you need to generate the gcloud credential file; you can do this by:
+
+- Downloading gcloud CLI (follow instructions on https://cloud.google.com/sdk/docs/install)
+- Running `gcloud auth appplication-default login` in the terminal
+
+Then you need to copy the credential file that was added to a fille called
```
dlatypov: nit: s/fille/file

README.md
```diff
@@ -37,6 +37,16 @@ Then you need to copy the cookies that were added to a file called
 
 NOTE: YOU MUST ADD A COMMENT TO THE TOP OF THE COOKIE FILE: `# HTTP Cookie File`
 
+### Setting up Google Cloud credentials
+
+First you need to generate the gcloud credential file; you can do this by:
+
+- Downloading gcloud CLI (follow instructions on https://cloud.google.com/sdk/docs/install)
```
dlatypov: very nit: like in the previous PR, can we switch this too to imperative tense? ("Downloading" -> "Download")

Dockerfile
```diff
@@ -1,12 +1,18 @@
 # set base image (host OS)
 FROM python:3.8
 
+# allow Google API to utilize dowloaded credentials
```
dlatypov: s/dowloaded/downloaded

(you fixed this already, but the fix ended up in the child commit on accident)

src/git.py
```diff
@@ -155,13 +156,13 @@ def _cleanup_git_dir(self) -> None:
         shutil.rmtree(self._git_dir)
 
     # Pass in the dao so that patches can be updated when they are pushed, this way less lost data when an error happens
-    def apply_patchset_and_cleanup(self, patchset: Patchset, messages_dao: message_dao.MessageDao):
+    def apply_patchset_and_cleanup(self, patchset: Patchset, local_mapping: Dict[str, Message], messages_dao : message_dao.MessageDao):
```
bjh83: nit: the space introduced after `messages_dao` is extraneous.

src/message_dao.py
```diff
@@ -12,27 +12,94 @@
...
+            )
+            # Mapping from name of state attribute to 
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS States"
+                "(state_name VARCHAR(255) NOT NULL,"
+                "value VARCHAR(255) NOT NULL,"
+                "PRIMARY KEY (state_name))"
+            )
+        self.connection.commit()
+        self.local_mapping = self._get_all()
```
bjh83: What's the purpose of the `local_mapping`? Is it supposed to be a cache?

The hash map that this replaces is already getting pretty big, why don't we just access the database directly? Are the database accesses that this sits in front of too expensive?

src/git.py
```diff
@@ -155,13 +156,13 @@ def _cleanup_git_dir(self) -> None:
...
         if not os.path.isdir(self._git_dir):
             self._setup_git_dir()
         for patch in patchset.patches:
             # This should cause the server to clean up the git dir because of potential failures
             try:
-                message = messages_dao.get(patch.message_id)
+                message = local_mapping.get(patch.message_id)
```
bjh83: Why are we replacing the DAO abstraction with a raw hash_map? If the DAO abstraction doesn't work, I would rather add another layer of abstraction rather than expose a raw hash map; even so, I am not sure that's the case here.

src/message_dao.py
```diff
@@ -12,27 +12,94 @@
...
+
+import pymysql
+import pickle
 from message import Message
+from google.cloud.sql.connector import Connector
+from dotenv import load_dotenv
 
 EPOCH_HASH = 'ae9e7be4a03765456fe38287533e6446e8bbc93c'
 
 class MessageDao(object):
```
bjh83, line 29: Note: For the tests, you will probably want to make a fake MessageDao that works like the old one so that the tests don't have to spin-up a database.

bjh83:
Can you rebase the above Dockerfile?

src/message_dao.py
dlatypov: I talked with Lenhard about this offline.

The idea was to not commit messages to the DB until after we uploaded them.
And so to make the `find()` function work as before, we needed some way of tracking "found-but-not-yet-uploaded" messages.

I suggested that we get rid of `find()` entirely, eliminating the need for `local_mapping`.
The places that use it can just use the Message objects they already had access to instead of going through the archive object.

src/git.py
```diff
@@ -155,16 +155,14 @@ def _cleanup_git_dir(self) -> None:
         shutil.rmtree(self._git_dir)
 
     # Pass in the dao so that patches can be updated when they are pushed, this way less lost data when an error happens
-    def apply_patchset_and_cleanup(self, patchset: Patchset, messages_dao: message_dao.MessageDao):
+    def apply_patchset_and_cleanup(self, patchset: Patchset, message: Message, message_dao: message_dao.MessageDao):
```
bjh83, line 159: I don't think I understand how this new message parameter works. Can you add a comment to this function that explains what the parameters do?

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+                "value VARCHAR(255) NOT NULL,"
+                "PRIMARY KEY (state_name))"
+            )
+        self.connection.commit()
+
+
+    def store(self, message: Message) -> None:
+        # Convert children into list of ids and store message
+        json_content = json.dumps(message.content)
+        comp_content = zlib.compress(json_content.encode())
```
bjh83: Is this necessary? How much space are we using? I would rather be able to look at the data in the database outside of the application.

bjh83 (changes requested):
Just a couple of minor comments.

src/git.py
lenhard-thomas: I'll add a comment, but the gist of it is that, previously, we'd utilize the patch's message id to look up the message in the database. However, that lookup is unnecessary as we already have these messages stored prior to calling this method. Thus, it's more efficient to simply pass these messages into the method.

src/message_dao.py
lenhard-thomas: On average, we're saving 1.5k bytes of data by compressing it. However, the largest size after compression is 21k bytes vs 97k bytes prior to compression. It's not inherently clear which approach is better.

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+                "CREATE TABLE IF NOT EXISTS States"
+                "(state_name VARCHAR(255) NOT NULL,"
+                "value VARCHAR(255) NOT NULL,"
+                "PRIMARY KEY (state_name))"
+            )
+        self.connection.commit()
+
+
+    def store(self, message: Message) -> None:
+        # Convert children into list of ids and store message
```
lenhard-thomas: Outdated comment; will be removed in next iteration.

src/archive_converter.py
```diff
@@ -21,55 +21,44 @@
...
-        return self._message_dao.size()
-
-    def find(self, message_id : str) -> Message:
-        message = self._message_dao.get(message_id)
-        if message is None:
-            raise ValueError(f'Could not find message: {message_id}')
-        return message
-
-    def _add_messages_to_index(self, new_messages : List[Message]) -> None:
+    def _add_messages_to_index(self, new_messages: Dict[str, Message]) -> None:
```
dlatypov: nit: can we rename this function now?
the name implies we might be mutating the current object by storing the new messages.

src/archive_converter.py
```diff
@@ -21,55 +21,44 @@
...
-                continue
-
-        for message in need_parent:
-            parent = self._message_dao.get(message.in_reply_to)
-            if not parent:
-                logging.info('Could not find parent email, dropping %s', message.debug_info())
-                continue
-            parent.children.append(message)
-            self._message_dao.store(parent)
+                if message.in_reply_to in new_messages:
```
dlatypov: nit: can we rewrite this as:

```py
if not message.in_reply_to:
   continue

# now we have less nesting
if message.in_reply_to in new_messages:
  ...
```

src/archive_converter.py
```diff
@@ -21,55 +21,44 @@
...
+                    parent = new_messages.get(message.in_reply_to)
+                    parent.children.append(message)
+                else:
+                    parent = self._message_dao.get(message.in_reply_to)
+                    if not parent:
+                        logging.info('Could not find parent email, dropping %s', message.debug_info())
+                        continue
+                    parent.children.append(message)
+                    new_messages[parent.id] = parent
```
dlatypov: presume we're adding the parent to the local map of new messages to ensure we store the updated message in the DB. (I'll have to read through the code some more to think if there's any adverse side effects of this)

This feels like something we should call out it in the docstring.

src/message.py
```diff
@@ -67,6 +67,17 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return str(self)
 
+    def __eq__(self, other) -> bool:
```
dlatypov, line 74: not actionable in this PR: hmm, maybe we should consider making Message a dataclass.
This `__eq__` definition is fine as-is.

But in the long-term, it would be nice to switch to an automatically generated one so we don't have bugs if fields change, etc.

https://docs.python.org/3/library/dataclasses.html

src/message.py
```diff
@@ -67,6 +67,17 @@ def __str__(self) -> str:
...
+        if not isinstance(other, Message):
+            return False
+        return (self.id == other.id and self.subject == other.subject 
+        and self.from_ == other.from_ and self.in_reply_to == other.in_reply_to 
+        and self.content == other.content and self.change_id == other.change_id
+        and self.archive_hash == other.archive_hash and self.children == other.children)
+
+    def __hash__(self) -> int:
+        return hash(self.id)
```
dlatypov: I assume this was to make `Set[Message]` work.
I'm sorta OK with this, but at the same time, it feels a bit wrong that `hash(m1) == hash(m2)` can be true while `m1 != m2` (e.g. if we didn't update `.children`).

I had initially suggested we convert `Set[Message]` to `Set[str]` (of ids) so we could specifically avoid doing something like this.
I'd still sorta prefer that.

dlatypov (commented):
I need to get back to this when I find some more time, publishing some comments for now.

This is looking a lot better. The use of dependency injection to clean up the unit tests particularly is nice to see.

src/message_dao_test.py
```diff
@@ -0,0 +1,96 @@
...
+from google.cloud.sql.connector import Connector
+
+import message_dao
+import archive_converter
+from test_helpers import test_data_path
+
+class MessageDaoTest(unittest.TestCase):
+
+    def setUp(self):
+        patcher = unittest.mock.patch.object(Connector, 'connect')
```
dlatypov: I tend to write such code as 

```python

from unittest import mock
...
  self.addCleanup(mock.patch.stopall)
  self.mock_connect = mock.patch.object(Connector, 'connect').start()
```

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+            raise Exception("Missing environment variable for name of database.")
+        with self.connection.cursor() as cursor:
+            cursor.execute("CREATE DATABASE IF NOT EXISTS " + db_name)
+            self.connection.select_db(db_name)
+            # Mapping from message id to message
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS Messages"
+                "(message_id VARCHAR(255) NOT NULL,"
+                "subject VARCHAR(255) NOT NULL,"
+                "from_ VARCHAR(255) NOT NULL,"
```
dlatypov, line 60: nit: I don't remember why we use `from_` as the field name, but I don't think we need to use that naming in the DB?

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+                "in_reply_to VARCHAR(255),"
+                "content BLOB NOT NULL,"
+                "archive_hash VARCHAR(255),"
+                "change_id VARCHAR(255),"
+                "PRIMARY KEY (message_id))"
+            )
+            # Mapping from name of state attribute to 
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS States"
```
dlatypov, line 69: so we're using this table as a generic key-value store?
Can we add a comment explaining that in this file somewhere?

We can either do it here or near the top of the file.
I thinking something like

```
We use two tables, Messages and States.
Message stores the messages we've uploaded and States is a key-value store which tracks things like "last_hash", the last Lore git commit we've processed.
```

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+            self.connection.select_db(db_name)
+            # Mapping from message id to message
+            cursor.execute(
+                "CREATE TABLE IF NOT EXISTS Messages"
+                "(message_id VARCHAR(255) NOT NULL,"
+                "subject VARCHAR(255) NOT NULL,"
+                "from_ VARCHAR(255) NOT NULL,"
+                "in_reply_to VARCHAR(255),"
+                "content BLOB NOT NULL,"
+                "archive_hash VARCHAR(255),"
```
dlatypov: Sorry, I should have brought this up a lot earlier, but I forgot.

more meta comment, the main thing we really need is a 
  `message_id` => `archive_hash`
mapping, and some indication of what message_id's we've uploaded.

With the `archive_hash` in hand, we can fetch the raw email from the archive.
From there we can get the subject, from, in_reply_to_, and content.
So we don't really need them to be in the DB necessarily.

The one other field we don't know from the `change_id`.
But we currently have this invariant: `change_id == sha1sum(msg_id)`.
For proper version tracking, we'll need to have this be `change_id == sha1sum(first_version_msg_id)`.

`first_version_msg_id` will be `== msg_id` for the first version of a patch (i.e. for v1's or for new patches introduced in a v2+ patch series)

So theoretically, all need to store is
```
  message_id,
  archive_hash,
  first_version_message_id,  # this might be the same as message_id for v1s, etc.
  in_reply_to,  # for easy lookups in _get_children()
``` 

There's some debuggability benefits to storing everything in one place, but something to consider.
It also has the drawback of you needing to do more to re-populate the message fields.

For concreteness, here's how we could fetch the message contents:
```
raw_email = subprocess.check_output(['git', '-C', archive_path, 'show', f'{hash}:m'])
msg = message.parse_message_from_str(raw_email, archive_hash=hash)
```

src/message_dao.py
```diff
@@ -12,27 +12,137 @@
...
+
+    def store(self, message: Message) -> None:
+        # Convert children into list of ids and store message
+        json_content = json.dumps(message.content)
+        comp_content = zlib.compress(json_content.encode())
+        sql = "REPLACE INTO Messages VALUES (%s, %s, %s, %s, %s, %s, %s)"
+        with self.connection.cursor() as cursor:
+            cursor.execute(sql, (message.id, message.subject, message.from_, 
+            message.in_reply_to, comp_content, message.archive_hash, message.change_id))
+        self.connection.commit()
```
dlatypov, line 85: note: for @lru_cache to be safe below in get() we should probably call 

```python
self.get.cache_clear()
```

Specifically, we need to invalidate the cache entry for the `in_reply_to` message bc we're adding a new child.

I don't know if functools provides a way to invalidate a specific cached value or not.
If you find a way to do that, then we should do that instead.

But I think just clearing the whole cache is OK enough.
If we care more, we'll probably have to write our own caching mechanism

src/git.py
lenhard-thomas: Added a comment to explain what the parameter does.

src/message_dao.py
lenhard-thomas: Removed content and instead added a lore link for ease of access.

src/archive_converter.py
lenhard-thomas: Renamed method to indicate that the purpose is to add children as needed.

src/archive_converter.py
lenhard-thomas: Done!

src/archive_converter.py
lenhard-thomas: Included this feature in the docstring.

src/message.py
lenhard-thomas: Left it as is for now. Might change it to a dataclass in a separate small PR.

src/message.py
lenhard-thomas: Removed hash due to the reasons you mentioned. Changed Set[Message] to Dict[str, Message] in order to minimize database lookups without dealing with hashing of messages.

src/message_dao_test.py
lenhard-thomas: Modified the code to look as such.

src/message_dao.py
lenhard-thomas: `from` is a keyword in python, which might be why it's named as such. I opted to maintain that naming scheme because `from` is also a keyword in SQL, so the two options are to avoid the keyword or to always use backticks when referring to the column. I opted for the former because it was already what we were doing within the python code.

src/message_dao.py
lenhard-thomas: Good idea. I added a docstring in the init of the DAO to explain the tables that are being created.

src/message_dao.py
lenhard-thomas: Changed the code to acquire messages using archive hash. Most columns ended up staying as they are needed for something other than just storing the message. However, by doing this we no longer have to store 'content', which is easily the largest column. For ease of access/replacement of column, I added a `lore_link` column that allows one to easily obtain the content of a particular row if needed.

src/message_dao.py
lenhard-thomas: Added a cache clear when storing a message that's a reply.

src/message_dao.py
dlatypov: Keeping the other fields SGTM.

Note: the lore link is just https://lore.kernel.org/linux-kselftest/<msg-id>
So we should probably drop it, tbh.

Even when we start archiving more, we can just use https://lore.kernel.org/lkml/<msg-id> to be agnostic of which sub-list it came from (and https://lore.kernel.org/all/ is an option as well)

src/message_dao.py
```diff
@@ -12,27 +12,141 @@
...
+                "(state_name VARCHAR(255) NOT NULL,"
+                "value VARCHAR(255) NOT NULL,"
+                "PRIMARY KEY (state_name))"
+            )
+        self.connection.commit()
+
+    def store(self, message: Message) -> None:
+        # Use only the portion of the subject that comes after the brackets that indicate patch information
+        reduced_subject = message.subject.partition("] ")[2].lower()
+        link = lore_link(message.id)
```
dlatypov, line 77: as noted in the other thread, we don't need to store the lore_link, it's just a concatenation of a constant and the msg_id.

src/message_dao.py
```diff
@@ -12,27 +12,141 @@
...
+    @lru_cache
+    def get(self, message_id: str) -> Optional[Message]:
+        sql = "SELECT archive_hash, change_id FROM Messages WHERE message_id=%s"
+        with self.connection.cursor() as cursor:
+            cursor.execute(sql, (message_id,))
+            res = cursor.fetchone()
+        if res is None:
+            return None
+        # Recreate the message object using the archive hash
+        archive_hash = res[0]
```
dlatypov: nit: I'd probably write this like

```python
archive_hash, change_id = res[0], res[1]
```

That way us grabbing out the change_id closer to the SELECT statement rather than 10 lines below (line 98 and line 108).

src/message_dao.py
bjh83: I actually suggested using the lore link. I like the idea because it makes it really easy to look the content of the patch up if we are accessing the database manually for research purposes.

src/message_dao_test.py
```diff
@@ -0,0 +1,96 @@
...
+        self.dao = message_dao.MessageDao('FAKE_GIT_PATH')
+
+    def assert_call_counts(self, connect, execute, commit):
+        self.assertEqual(connect, self.mock_connect.call_count)
+        self.assertEqual(execute, self.mock_cur.execute.call_count)
+        self.assertEqual(commit, self.mock_con.commit.call_count)
+
+    def test_init(self):
+        # Testing initialization of DAO
+        self.assert_call_counts(1, 3, 1)
```
dlatypov: optional: wdyt about
```python
self.assert_call_counts(connect=1, excecute=3, commit=1)
```

I think this would make the tests more readable.

I still think it's a bit tricky to follow the tests, since mocks and call counts always are.
If we don't want calls to be made, we should probably make them throw errors by default.

E.g. I'd prefer we have

```python
def setUp(self):
  ...
  self.assert_call_counts(1, 3, 1)
  
  self.mock_con.reset_mock() # reset the calls() tracking for all the mocks
  # we can maybe also set self.mock_connect.side_effect = RuntimeError("should not be called again after init!")
  ...

def test_store(self):
  ...
  self.mock_cur.assert_called_once_with(..., mock.ANY)  # maybe match the
  self.mock_commit.assert_called_once()
  
  # then call reset_mock() again
```

wdyt?

dlatypov (changes requested):
I still need to look over src/main.py a bit more closely and I had some comments about the use of mocking, but otherwise LGTM.

I think src/main.py is correct, but I just need some more time to wrap my head around everything going on it (not your fault, it was always a bit more complicated than it probably should be).

src/message_dao.py
lenhard-thomas: As Brendan mentioned, the idea was to have an easy to access link to the raw data itself. Although all we need is the message id, storing the link does allow for just a simple click. What are your thoughts?

src/message_dao.py
lenhard-thomas: Good call! Looks much cleaner that way.

src/message_dao_test.py
lenhard-thomas: I like the idea of explicit keyword argument to make the test more readable. Although, I'm not sure if we'll need it if we're instead throwing errors for mocks that shouldn't be run; `mock_connect` only runs on init and `mock_commit` only runs when storing. Since `mock_execute` is the only one that consistently shows up, it's probably better to just have a single `assertEquals(execute_count, mock_execute)`

src/message_dao.py
dlatypov: I don't have any strong preference.
Let's keep it if you and Brendan agree.

I don't personally think it'd be too useful, but it's only adding a small number of bytes.

src/message_dao_test.py
dlatypov: Sorry, I should have made it more clear.

The suggestions are on a spectrum:
* one end: annotate the parameters to make it a bit easier to follow
* other end: move away from doing asserts on call counts and switch to explicit `.assert_called_once_with(...)`. Rely on errors to enforce unexpected calls aren't made.

We can do a mix of both.

E.g. we could make it so connect() is called once and then would error out if called again.
Then we'd do `assert_call_counts(execute=1, commit=1)`, etc.

I'd personally prefer we go farther to "throw errors" end.

lenhard-thomas: Is there a benefit in checking the value of a mocked cursor (aka `self.mock_cur.assert_called_once_with(..., mock.ANY)`)? Wouldn't it just be copying the code under the `store` method?

bjh83: I agree with this. It makes it much clearer.

src/message_dao.py
bjh83: I am still leaning towards keeping the whole lore link: If we plan on opening this up to other email lists in the future, then we should store the list here as well. But then if we keep both the list name as well as the message id, then we are only going to save 23 chars. Now, I know this is 23 times "a lot", which is a lot; however, I think it also makes it meaningfully easier to look up the patch on lore then having to copy and paste and type 23 chars into chrome. For reference, a typical full lore link will be about 90 bytes.

I fully expect that we will spend non-trivial amounts of time on our DB researching improvements to the lkml-gerrit-bridge. Even if it is only interns in the summer, I think 1 intern next summer spending 10 hours messing around on the DB would make making the lore-link copy-and-paste-able worthwhile.

Now, there are other compromises we could make: We could just cut the "https://" off of the link, reducing the overhead of copy-and-paste-ability to 15 bytes. We could also make a go link ("go/llore" is available) to reduce the pain of typing a prefix before copying and pasting, but it's still enough characters that I would find typing it out to be annoying.

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+        mock_connection = self.mock_connect.return_value
+        self.mock_commit = mock_connection.commit
+        self.mock_cursor = mock_connection.cursor.return_value.__enter__.return_value
+        self.mock_execute = self.mock_cursor.execute
+        # Check initialization of DAO
+        self.dao = message_dao.MessageDao('FAKE_GIT_PATH')
+        self.mock_connect.assert_called_once()
+        self.mock_commit.assert_called_once()
+        self.assertEqual(3, self.mock_execute.call_count)
+        self.reset_mocks(return_value=True, side_effect=True)
```
dlatypov: hmm, I'm not sure we need to change the side_effect and return_value of the mocks.
They're not set to anything (besides other mocks).

So I think we can remove those parameters here and from the method itself

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+        # Testing storing Message that's not already in DAO 
+        email = archive_converter.generate_email_from_file(test_data_path('patch6.txt'))
+        sql_text = "REPLACE INTO Messages VALUES (%s, %s, %s, %s, %s, %s, %s)"
+        self.dao.store(email)
+        self.mock_execute.assert_called_once_with(sql_text, mock.ANY)
+        self.mock_commit.assert_called_once()
+        self.reset_mocks()
+
+        # Testing storing Message that's already in DAO 
+        email = archive_converter.generate_email_from_file(test_data_path('patch6.txt'))
```
dlatypov: this extra test case doesn't do anything, right?

i.e. it doesn't test any new codepaths or different state, since it's all mocked out.
we should drop this block and the reset_mocks() call above it.

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+    @mock.patch.object(message_dao, 'parse_message_from_str')
+    def test_get(self, mock_parse_msg, mock_check_output):
+        # Testing searching for a message id that's not in DAO
+        self.mock_commit.side_effect = RuntimeError("Shouldn't be called during get")
+        email = archive_converter.generate_email_from_file(test_data_path('patch6.txt'))
+        mock_parse_msg.return_value = email
+        self.mock_cursor.fetchone.return_value = (email.archive_hash, email.change_id)
+        self.assertEqual(email, self.dao.get('fake_message_id'))
+        self.assertEqual(2, self.mock_execute.call_count) 
+        self.reset_mocks()
```
dlatypov: this second part should be a separate test case.
In general, if you have to do stuff like reset_mocks() in a test case, that probably means you're testing something else and that belongs in a separate case.

Here, we want to do that.
```python
def test_get_missing(self):  # note: we don't need or want to mock check_output!
  ...
```

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Testing size of one dao
+        self.mock_cursor.fetchone.return_value = (1,)
+        self.assertEqual(1, self.dao.size())
+        self.mock_execute.assert_called_once()
+        self.reset_mocks()
+
+        # Testing size of many dao
+        self.mock_cursor.fetchone.return_value = (10,)
```
dlatypov: as noted before, we only need one of these blocks.

Since we're using mocks, the code path covered by all of them is the exact same.
While the underlying sql lib might do something different if the table is empty, our code doesn't.

We just return what the mock gives us.

Let's truncate this test func down into just one block

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+        self.mock_execute.assert_called_once()
+
+    def test_store_hash(self):
+        # Testing storing hash when it's not already in State table
+        self.dao.store_last_hash("some_hash")
+        sql_text = "REPLACE INTO States VALUES (%s, %s)"
+        self.mock_execute.assert_called_once_with(sql_text, mock.ANY)
+        self.mock_commit.assert_called_once()
+        self.reset_mocks()
+
```
dlatypov, line 72: ditto from get() above.
Since we're using mocks, nothing in our code behaves different if the message is in there or not.

Drop the second block and the `reset_mocks()` call above.

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+    def test_get_hash(self):
+        # Testing searching for hash when it's not already in State table
+        self.mock_commit.side_effect = RuntimeError("Shouldn't be called during get_hash")
+        self.mock_cursor.fetchone.return_value = None
+        self.assertEqual(message_dao.EPOCH_HASH, self.dao.get_last_hash())
+        self.mock_execute.assert_called_once()
+        self.reset_mocks()
+
+        # Testing searching for hash when it's already in State table
```
dlatypov: ditto from above, I'd make this a separate pair of test cases.
 `test_get_hash` and `test_get_hash_missing` or something like that.

src/message_dao_test.py
```diff
@@ -0,0 +1,115 @@
...
+        self.mock_cursor.fetchone.return_value = ("fake_hash",)
+        self.assertEqual("fake_hash", self.dao.get_last_hash())
+        self.mock_execute.assert_called_once()
+
+if __name__ == '__main__':
+    # TODO(@lenhard): Issue with Google's Connector that causes segmentation fault
```
dlatypov: nit: we usually would write TODO(email), so TODO(lenhard@google.com)

dlatypov (changes requested):
LGTM, just some nits about the unit test file and I think we're good to go!

src/message_dao_test.py
lenhard-thomas: Good point. Removed the extra parameters.

src/message_dao_test.py
lenhard-thomas: Removed extra test case since code coverage does remain the same

src/message_dao_test.py
lenhard-thomas: Separated the test cases.

src/message_dao_test.py
lenhard-thomas: Yep, code coverage remains the same with just one block; removed the other two test cases.

src/message_dao_test.py
lenhard-thomas: Removed the extra test case.

src/message_dao_test.py
lenhard-thomas: Split the test case into two unit tests.

src/message_dao_test.py
lenhard-thomas: Got it; I was just going off of the TODO in main.py.

src/message_dao_test.py
```diff
@@ -0,0 +1,79 @@
...
+        mock_connection = self.mock_connect.return_value
+        self.mock_commit = mock_connection.commit
+        self.mock_cursor = mock_connection.cursor.return_value.__enter__.return_value
+        self.mock_execute = self.mock_cursor.execute
+        # Check initialization of DAO
+        self.dao = message_dao.MessageDao('FAKE_GIT_PATH')
+        self.mock_connect.assert_called_once()
+        self.mock_commit.assert_called_once()
+        self.assertEqual(3, self.mock_execute.call_count)
+        self.mock_connect.reset_mock()
```
dlatypov: very nit: we probably don't need to reset the connect mock since we're overriding side effect to raise an error

src/message_dao_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+    @mock.patch.object(subprocess, 'check_output')
+    @mock.patch.object(message_dao, 'parse_message_from_str')
+    def test_get(self, mock_parse_msg, mock_check_output):
+        self.mock_commit.side_effect = RuntimeError("Shouldn't be called during get")
+        email = archive_converter.generate_email_from_file(test_data_path('patch6.txt'))
+        mock_parse_msg.return_value = email
+        self.mock_cursor.fetchone.return_value = (email.archive_hash, email.change_id)
+        self.assertEqual(email, self.dao.get('fake_message_id'))
+        self.assertEqual(2, self.mock_execute.call_count)
```
dlatypov: I already forgot and got confused why this would be 2.
We could add a comment explaining _get_children() is called to.

Or if we want, we could also do

```python
class StrContains(string):
   def __eq__(self, other):
      return self in other
...

self.assertEqual(2, self.mock_execute.call_count)
self.mock_execute.assert_has_calls([
  mock.call(StrContains("where message_id=%s"), mock.ANY),  # get this msg
  mock.call(StrContains("where in_reply_to=%s"), mock.ANY),  # get children
 ])
 ```

File: src/message_dao_test.py [hunk 33]
```diff
@@ -0,0 +1,79 @@
...
+        self.mock_execute.assert_called_once()
+
+    def test_store_hash(self):
+        self.dao.store_last_hash("some_hash")
+        sql_text = "REPLACE INTO States VALUES (%s, %s)"
+        self.mock_execute.assert_called_once_with(sql_text, mock.ANY)
+        self.mock_commit.assert_called_once()
+
+    def test_get_hash(self):
+        self.mock_cursor.fetchone.return_value = ("fake_hash",)
```
dlatypov, line 75: nit: also make self.mock_commit.side_effect an error here?

dlatypov (commented):
Last final round of nits, looks good!

src/message_dao_test.py
lenhard-thomas: Yeah, it's unnecessary so it's been removed to improve readability.

src/message_dao_test.py
lenhard-thomas: Clever solution! Modified to utilize StrContains in addition to asserting call counts: that way it's explicit about where the execute methods calls are coming from.

src/message_dao_test.py
lenhard-thomas: Added the runtime error there as well.

dlatypov (approved):

PR #50, opened 2022-08-08
Patch associator

lenhard-thomas:


src/main.py
```diff
@@ -112,7 +114,8 @@ def update_convert_upload(self) -> None:
                     parent = self.message_dao.get(message.in_reply_to)
                 messages_with_new_comments[parent.id] = parent
 
-        self.upload_messages(messages_to_upload)
+        # Sort messages so that older versions are uploaded first
+        messages_to_upload.sort(key=lambda x : x.subject)
```
dlatypov: Note: this won't work if you have `PATCH v9` and `PATCH v10`

src/message_dao.py
```diff
@@ -109,6 +109,15 @@ def get(self, message_id: str) -> Optional[Message]:
         msg.children = self._get_children(message_id)
         return msg
 
+    @lru_cache()
+    def previous_version_candidates(self, message: Message) -> List[Message]:
+        reduced_subject = message.subject.partition("] ")[2].lower()
```
dlatypov: can we make this a method on `Message` given we have the same code elsewhere?

Also, perhaps we should call it "normalize_subject" or something like that given we're also lowercasing it.

src/patch_associator.py
```diff
@@ -0,0 +1,60 @@
...
+        candidates = self._message_dao.previous_version_candidates(message)
+        candidates = self.newest_first(candidates)
+        previous_version_number = message.version() - 1
+        compiled = re.compile(fr'\[.+ v{previous_version_number}.*\]')
+        for candidate in candidates:
+            if compiled.match(candidate.subject):
+                return candidate
+            if previous_version_number == 1:
+                # Consider subjects that don't have a version number
+                compiled_no_version = re.compile(fr"\[PATCH\]")
```
dlatypov: nit just `r""`, we don't need `f` in this one

src/patch_associator.py
```diff
@@ -0,0 +1,60 @@
...
+import re
+import subprocess
+
+from typing import Dict, List, Optional
+from message import Message
+from patch_parser import Patch
+from message_dao import MessageDao
+
+class PatchAssociator(object, metaclass = abc.ABCMeta):
+    def __init__(self, message_dao : MessageDao) -> None:
```
dlatypov: note: perhaps `message_dao` shouldn't be a member, but a parameter to `get_previous_version()`?

Unless we're planning on making a stateful PatchAssociator that depends specifically on one `MessageDao`, we don't need to store it here.

More variables = more changes for bugs.
E.g. what if the server and patch associator accidentally end up w/ different instances?
It could happen in tests accidentally, etc.

src/patch_associator.py
```diff
@@ -0,0 +1,60 @@
...
+        candidates_with_time = [(message, self.get_time(message)) for message in candidates]
+        # sort by newest commit to latest commit
+        candidates_with_time.sort(key=lambda x: -x[1])
+        return [message for (message, time) in candidates_with_time]
+
+    def get_previous_version(self, message: Message) -> Optional[Message]:
+        candidates = self._message_dao.previous_version_candidates(message)
+        candidates = self.newest_first(candidates)
+        previous_version_number = message.version() - 1
+        compiled = re.compile(fr'\[.+ v{previous_version_number}.*\]')
```
dlatypov: nit: we're not requiring "[PATCH v\d+]" in the regex here, but we are below for v1.
I think we should update the regex here to require PATCH as well.

src/message_dao.py
```diff
@@ -109,6 +109,15 @@ def get(self, message_id: str) -> Optional[Message]:
...
+        sql = "SELECT message_id FROM Messages WHERE from_=%s AND reduced_subject=%s"
```
dlatypov: note: should we require a non-empty change id?
Since we only really care about the previous versions for the change id anyway.

And it would guard against some edge cases:
What happens if someone replies to a patch but they don't change the subject to start with "Re: "?
Or what happens if we just fail to apply the parent patch (since it doesn't apply to torvalds/master), but the child does? (Answer, we could publish v2 as a patchset 1, which could be confusing...)
Or what happens if they send a malformed buggy version of a patch and then resend again with the same subject and version #?

src/message_dao.py
```diff
@@ -150,3 +159,6 @@ def store_last_hash(self, last_hash: str) -> None:
 
     def get_last_hash(self) -> int:
         return self.last_hash
+
+    def previous_versions(self, message: Message) -> List[Message]:
```
dlatypov: should this be called previous_version_candidates() as well?

src/message_dao.py
```diff
@@ -109,6 +109,15 @@ def get(self, message_id: str) -> Optional[Message]:
         msg.children = self._get_children(message_id)
         return msg
 
+    @lru_cache()
```
dlatypov: hmm, what benefit does caching give us?
I don't imagine we'd call this function multiple times for the same exact message.

src/message_dao.py
```diff
@@ -109,6 +109,15 @@ def get(self, message_id: str) -> Optional[Message]:
...
+    def previous_version_candidates(self, message: Message) -> List[Message]:
```
dlatypov: bigger question: we've made PatchAssociator be an ABC, but the core association logic is here, in the MessageDao.

I'm not sure we can meaningfully make new implementations of the PatchAssociator on its own.

I don't know how we want to handle this, since we don't want to expose SQL to the associator, necessarily.

We could make this method be like `find_with_matching_subject(self, reduced_subject: str, from_: str)`, which is something we can implement in the fake dao...

bjh83: We could make another class called a PatchCandidateSelector (or something like that) which could be passed into the MessageDao from the PatchAssociator. The PatchCandidateSelector could effectively implement this method and would be tied to a PatchAssociator. Thoughts?

src/message_dao.py
bjh83: People do *sometimes* upload a new revision within hours of an earlier revision, but I am with @dlatypov, I don't think that happens frequently enough to pay for the overhead of an lru_cache, let alone make it worthwhile.

src/patch_associator.py
bjh83: +1

Although we do see patches based on RFCs, that versioning logic isn't covered here. I think for now we can require that it says PATCH.

src/main.py
lenhard-thomas: Changed sorting to be based on the version number directly.

src/message_dao.py
lenhard-thomas: Changed it to a method within the `Message` class and also created a new attribute named `normalized_subject`.

src/patch_associator.py
lenhard-thomas: Yep! Changed accordingly.

src/patch_associator.py
lenhard-thomas: Moved message_dao to a parameter of the `get_previous_version` method.

src/git.py
```diff
@@ -23,6 +23,8 @@
 from message import lore_link, Message
 from patch_parser import Patch, Patchset
 from absl import logging
+from typing import Optional
+from patch_associator import PatchAssociator, SimplePatchAssociator
```
dlatypov: nit: looks like `SimplePatchAssociator` is not used in this file

src/patch_associator.py
```diff
@@ -0,0 +1,61 @@
+# Copyright 2020 Google LLC
```
dlatypov: nit: 2022

src/patch_associator.py
```diff
@@ -0,0 +1,61 @@
...
+        # sort by newest commit to latest commit
+        candidates_with_time.sort(key=lambda x: -x[1])
+        return [message for (message, time) in candidates_with_time]
+
+    def get_previous_version(self, message: Message, message_dao: MessageDao) -> Optional[Message]:
+        candidates = message_dao.find_matching(
+                     normalized_subject = message.normalized_subject, 
+                     from_ = message.from_)
+        candidates = self._newest_first(candidates)
+        previous_version_number = message.version() - 1
```
dlatypov: we shouldn't see a `PATCH v0` or even `PATCH v1`, so this could instead be something like

```python
if message.version() == 2:
  regex = re.compile(r'\[PATCH(?!.*v.*]).*]'))
else:
  prev_version = message.version() - 1
  regex = re.compile(fr'\[PATCH v{prev_version}.*\]')
```

src/patch_associator.py
```diff
@@ -0,0 +1,61 @@
...
+        with_version = re.compile(fr'\[PATCH v{previous_version_number}.*\]')
```
dlatypov: tangent: our regexes are allowing stuff like `[PATCH v8 random text]`.
I don't think we need to?

Our regexes would be simpler as `r'\[PATCH\]'`

src/message_dao.py
lenhard-thomas: Query now excludes messages where the change id is empty. Alternate solution to consider might be excluding the existence of null change ids within the database; that is, not allowing the `change_id` column to be a null value.

src/message_dao.py
lenhard-thomas: Whoops. Left over from the previous naming scheme. Changed accordingly.

src/patch_associator.py
```diff
@@ -0,0 +1,61 @@
...
+            ['git', '-C', self.git_path , 'show', '-s', '--format=%ct', message.archive_hash]))
+
+    def _newest_first(self, candidates: List[Message]):
+        candidates_with_time = [(message, self._get_time(message)) for message in candidates]
+        # sort by newest commit to latest commit
+        candidates_with_time.sort(key=lambda x: -x[1])
+        return [message for (message, time) in candidates_with_time]
+
+    def get_previous_version(self, message: Message, message_dao: MessageDao) -> Optional[Message]:
+        candidates = message_dao.find_matching(
```
dlatypov, line 50:
```python
if message.version() < 2:
  return None
```

Shouldn't we do this? If it's not a v2 or above, there shouldn't be a previous version.
(Not unless we start looking for RFC versions of patches, etc.)

src/message_dao.py
lenhard-thomas: That's true, I can't imagine many cases where cache helps out so I've removed it.

src/message_dao.py
lenhard-thomas: Incorporated the `find_with_matching_subject` idea but made it more general (as discussed in 1:1 w/ Daniel). Now allows for expansion of matching to other attributes with quite ease.

src/message_dao.py
```diff
@@ -109,6 +108,26 @@ def get(self, message_id: str) -> Optional[Message]:
...
+        
+        clauses = [f' {attr}=%s' for attr, _ in non_empty]
+        values = [value for _, value in non_empty]
+
+        query = "SELECT message_id FROM Messages WHERE change_id IS NOT NULL AND" + " AND".join(clauses)
+
+        with self.connection.cursor() as cursor:
+            cursor.execute(query, tuple(values))
+            res = cursor.fetchall()
+        return [self.get(message_id) for message_id in (tup[0] for tup in res)]
```
dlatypov: we have a double comprehension going on.

Would it be simpler as 
```python
[self.get(tup[0]) for tup in res]
```

Alternatively, we could do
```python
[self.get(msg_id) for (msg_id,) in res]
```

but I think explicitly doing `tup[0]` is clearer than unpacking a single-element tuple

src/message_dao.py
```diff
@@ -150,3 +169,15 @@ def store_last_hash(self, last_hash: str) -> None:
...
         return self.last_hash
+
+    def find_matching(self, normalized_subject: str = "", from_: str = "") -> List[Message]:
+        criteria = {
+            "normalized_subject": normalized_subject,
+            "from_": from_,
+        }
+    
+        def _Match(msg: Message):
+            return all(getattr(msg, attr) == value and msg.change_id is not None for attr, value in criteria.items() if value != "")
```
dlatypov: we're evaluating `msg.change_id is not None` for every attribute we match on.

Perhaps instead
```python
def _Match(msg: Message):
  if msg.change_id is None:
     return False
   return all(getattr...)
```

dlatypov (changes requested):
Nice, I just have a few nits.

src/patch_associator.py
lenhard-thomas: There are some patches in the form of `PATCH V1`, but we could certainly optimize the case of the message.version being equal to one in order to return early.

src/patch_associator.py
lenhard-thomas: Do you mean that rather than looking for the previous version, just always look for the first version?

src/patch_associator.py
lenhard-thomas: Yep, an early return can save on a bunch of database access.

src/message_dao.py
lenhard-thomas: That'd certainly be simpler. I was trying to do it the 2nd way for clearer method calls, but I didn't realize we could unpack directly like that. I'll stick with the first way that you suggested.

src/patch_associator.py
dlatypov: Ah, I meant we should look for either
  [PATCH]
  [PATCH v\d+]
We don't need to support text after the version number like ew do right now

lenhard-thomas: I see. I could've been more explicit about it with my regex, but the idea was to include patches featuring the index of the patchset (e.g `[PATCH v3 1/6]`) as well as patches with more information than normal (e.g `[PATCH net v4 1/3]` and `[PATCH v4 net-next 0/6]`). What are your thoughts?

dlatypov: Oh brain fart, I forgot about the 1/6 bit.
Please ignore this whole thread.

src/message_dao.py
lenhard-thomas: Makes sense; no need to check it per attribute when the return is dependent on just the message.

